PR #36, opened 2026-03-24
feat: Phase 5c — Admin Portal & Member Management

### LINKED ISSUE #14: Phase 5c: Admin Portal & User Management
## Description
Implement admin portal and tenant-level user management.

**Descope features:** Admin portal, widgets, user management

## Tasks
- [ ] Embedded Descope admin portal for tenant admins
- [ ] User management within tenants:
  - Invite users by email
  - Assign/change roles
  - Deactivate users
  - View login history
- [ ] Admin portal configuration via Terraform (`admin_portal` block)
- [ ] Widget configuration for admin UI
- [ ] Member list with role badges and status indicators

## Acceptance Criteria
- Tenant admins can manage their team members
- Invite flow sends email and assigns role on acceptance
- Deactivated users cannot access the tenant
- Admin portal is only accessible to users with `admin` or `owner` role

## Dependencies
- #5, #6

## Phase
Phase 5: Flows, Connectors & Admin

jamescrowley321:
## Summary
- Add member management API: list, invite, deactivate, activate, remove users
- Frontend: MemberManagement page with invite form, role badges, status toggle
- All operations require admin/owner role and tenant context
- 143 unit tests

## Test plan
- [x] 143 unit tests pass (lint + pytest)
- [x] TypeScript compiles clean
- [x] Viewer cannot manage members (403)
- [x] Default role assignment on invite (member)
- [ ] Manual test with live Descope project

Automated PR from ralph loop task T26. Closes #14

### DIFF
--- a/backend/app/services/descope.py
+++ b/backend/app/services/descope.py
@@ -117,6 +117,71 @@ class DescopeManagementClient:
             {"id": tenant_id, "customAttributes": custom_attributes},
         )
 
+    async def create_access_key(
+        self,
+        name: str,
+        tenant_id: str,
+        expire_time: int | None = None,
+        role_names: list[str] | None = None,
+    ) -> dict:
+        """Create an access key scoped to a tenant."""
+        body: dict = {"name": name, "tenantId": tenant_id}
+        if expire_time is not None:
+            body["expireTime"] = expire_time
+        if role_names:
+            body["roleNames"] = role_names
+        resp = await self._request("/v1/mgmt/accesskey/create", body)
+        return resp.json()
+
+    async def search_access_keys(self, tenant_id: str) -> list[dict]:
+        """List access keys filtered by tenant."""
+        resp = await self._request("/v1/mgmt/accesskey/search", {"tenantIds": [tenant_id]})
+        return resp.json().get("keys", [])
+
+    async def load_access_key(self, key_id: str) -> dict:
+        """Load a single access key by ID."""
+        resp = await self._request("/v1/mgmt/accesskey/load", {"id": key_id})
+        return resp.json().get("key", {})
+
+    async def deactivate_access_key(self, key_id: str) -> None:
+        """Deactivate (revoke) an access key."""
+        await self._request("/v1/mgmt/accesskey/deactivate", {"id": key_id})
+
+    async def activate_access_key(self, key_id: str) -> None:
+        """Reactivate a previously deactivated access key."""
+        await self._request("/v1/mgmt/accesskey/activate", {"id": key_id})
+
+    async def delete_access_key(self, key_id: str) -> None:
+        """Permanently delete an access key."""
+        await self._request("/v1/mgmt/accesskey/delete", {"id": key_id})
+
+    async def invite_user(self, email: str, tenant_id: str, role_names: list[str] | None = None) -> dict:
+        """Create a user and assign them to a tenant with roles."""
+        tenants = [{"tenantId": tenant_id}]
+        if role_names:
+            tenants[0]["roleNames"] = role_names
+        resp = await self._request(
+            "/v1/mgmt/user/create",
+            {"loginId": email, "email": email, "tenants": tenants},
+        )
+        return resp.json().get("user", {})
+
+    async def search_tenant_users(self, tenant_id: str) -> list[dict]:
+        """Search for users belonging to a specific tenant."""
+        resp = await self._request("/v1/mgmt/user/search", {"tenantIds": [tenant_id]})
+        return resp.json().get("users", [])
+
+    async def update_user_status(self, user_id: str, status: Literal["enabled", "disabled"]) -> None:
+        """Update user status."""
+        await self._request("/v1/mgmt/user/updateStatus", {"loginId": user_id, "status": status})
+
+    async def remove_user_from_tenant(self, user_id: str, tenant_id: str) -> None:
+        """Remove a user from a specific tenant (does not delete the user globally)."""
+        await self._request(
+            "/v1/mgmt/user/update/tenant/remove",
+            {"loginId": user_id, "tenantId": tenant_id},
+        )
+
     async def close(self) -> None:
         """Close the underlying HTTP client if we own it."""
         if self._http_client is not None and not self._owns_client:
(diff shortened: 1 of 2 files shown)


jamescrowley321:
## Adversarial Code Review — PR #36

### MUST FIX

**1. Member operations (deactivate/activate/remove) have no tenant-scoping — any admin can operate on any user globally**
`backend/app/routers/users.py` — The `deactivate_member`, `activate_member`, and `remove_member` endpoints accept a `user_id` path parameter and call the Descope Management API directly (`update_user_status`, `delete_user`) without verifying that the target user actually belongs to the caller's tenant. The `_tenant_id` dependency is injected but discarded (prefixed with `_`). An admin in Tenant A can deactivate or permanently delete a user who belongs only to Tenant B by guessing/enumerating user IDs. This is a **cross-tenant privilege escalation**.

Recommended fix: call `search_tenant_users(tenant_id)` or `load_user(user_id)` and verify the target user has a matching tenant membership before performing the action.

**2. `create_tenant` has no authorization check — any authenticated user can create tenants**
`backend/app/routers/tenants.py` line `create_tenant` — The endpoint only requires `get_claims` (any authenticated user). There is no RBAC check. Any logged-in user, including a `viewer`, can create unlimited tenants via the Descope Management API. This is likely unintended for a SaaS application.

**3. `get_descope_client()` creates a new HTTP client on every single API call**
`backend/app/services/descope.py` — Every method creates a new `httpx.AsyncClient()` via `async with`, which means a new TCP connection + TLS handshake per request. In a production SaaS app under load, this will exhaust file descriptors and degrade latency. The client should be a singleton or use connection pooling. Additionally, `get_descope_client()` itself is called on every request, re-reading env vars each time — it should be cached.

**4. `_verify_key_tenant` uses a separate `get_descope_client()` call from the router**
`backend/app/routers/accesskeys.py` — `_verify_key_tenant` calls `get_descope_client()` internally, then the router handler calls `get_descope_client()` again for the actual operation. Since each call constructs a new client, this doubles the overhead. More importantly, there is a TOCTOU race: the key could be reassigned between the verification call and the action call. These should use the same client instance and ideally be atomic.

### SHOULD FIX

**5. `update_user_status` accepts arbitrary strings — no validation of `status` parameter**
`backend/app/services/descope.py` line `update_user_status` — The `status` parameter is typed as `str` but the Descope API only accepts `"enabled"` or `"disabled"`. The docstring mentions this but there is no runtime validation. A programming error could send invalid status values silently.

**6. `invite_user` has no email format validation**
`backend/app/routers/users.py` — `InviteUserRequest.email` is typed as `str` but has no email validation (e.g., `EmailStr` from pydantic). Invalid email strings will be sent to the Descope API which may reject them with an opaque error.

**7. Bare `except Exception` blocks swallow errors silently in multiple places**
`backend/app/routers/attributes.py` — `get_profile` and `get_tenant_settings` catch all exceptions and return empty data. This makes debugging production issues very difficult. At minimum, log the exception. The same pattern appears in `tenants.py` `get_current_tenant`.

**8. No pagination on `search_tenant_users` and `search_access_keys`**
`backend/app/services/descope.py` — Both search methods send requests without limit/offset parameters. For tenants with many users or keys, the Descope API may return truncated results or the response may be very large. The list endpoints should support pagination.

**9. `RoleAssignmentRequest` accepts a `tenant_id` field from the client — unnecessary attack surface**
`backend/app/routers/roles.py` — The request body includes `tenant_id` which is then checked against `current_tenant`. Since it must always equal `current_tenant` or be rejected, it should not be a client-supplied field at all. The tenant should be derived from the JWT `dct` claim exclusively.

**10. SQLite default database path `./data.db` is relative and environment-dependent**
`backend/app/models/database.py` — The default `sqlite:///./data.db` resolves relative to the CWD at startup, which varies between dev, Docker, and systemd. This could cause data loss if the app starts from a different directory. Should use an absolute path or require explicit configuration.

**11. Frontend member management has no confirmation dialog for destructive actions**
`frontend/src/pages/MemberManagement.tsx` — The "Remove" button permanently deletes a user with no confirmation. Same issue in `AccessKeys.tsx` for key deletion. Users can accidentally destroy data with a single click.

**12. Frontend `handleToggleStatus` and `handleRemove` swallow fetch errors**
`frontend/src/pages/MemberManagement.tsx` — `handleToggleStatus` and `handleRemove` use bare `await apiFetch(...)` without checking `res.ok` or catching errors. If the API call fails, the UI silently calls `loadMembers()` anyway, making it appear as if nothing happened (or the stale data remains).

**13. `navigator.clipboard.writeText` has no error handling and fails silently in non-HTTPS contexts**
`frontend/src/pages/AccessKeys.tsx` — The clipboard API requires a secure context (HTTPS or localhost). In some deployment scenarios, this will throw, but there is no catch block. The `copied` state will remain false with no user feedback about the failure.

**14. Admin can assign arbitrary role names — no allowlist validation**
`backend/app/routers/roles.py` — `RoleAssignmentRequest.role_names` accepts any string list. An admin could attempt to assign nonexistent roles, which may cause unexpected behavior in Descope. Should validate against the known role set (`owner`, `admin`, `member`, `viewer`).

### NITPICK

**15. `TenantResource.created_at` is stored as `str` instead of a proper datetime type**
`backend/app/models/tenant.py` — Using ISO format strings works but prevents database-level date queries and sorting. SQLModel supports `datetime` fields natively.

**16. Inconsistent error response format**
Some endpoints return `{"detail": "..."}` (via HTTPException), while success responses use varied shapes (`{"status": "...", "key_id": "..."}`, `{"members": [...]}`, etc.). Consider a standard envelope.

**17. Frontend `useApiClient` uses relative fetch paths — assumes same-origin deployment**
`frontend/src/hooks/useApiClient.ts` — `fetch(path, ...)` with paths like `/api/keys` will fail if the frontend and backend are on different origins (common in dev). This is pre-existing but worth noting since the PR adds many new API calls depending on it.

**18. Test fixtures are duplicated across 7 test files**
All test files define their own `ADMIN_CLAIMS`, `VIEWER_CLAIMS`, `client` fixture, and `_set_env` fixture. Extract these into a shared `conftest.py` to reduce duplication.

**19. `accesskeys.py` router uses `_admin_roles` (underscore prefix) — intent is unclear**
The underscore prefix conventionally indicates an unused variable, but the dependency is not unused (it enforces the RBAC check as a side effect). This is misleading.

---

**Summary**: The most critical issues are (1) the cross-tenant user operation vulnerability in the members router, (2) the missing authorization on tenant creation, and (3) the per-request HTTP client creation in the Descope service. The rest are defense-in-depth gaps and DX improvements.

PR #37, opened 2026-03-24
feat: add security headers middleware

### LINKED ISSUE #28: Add security headers middleware (HSTS, X-Frame-Options, CSP)
## Goal

Add security headers middleware to the FastAPI backend for defense-in-depth.

## Context

As an auth-focused SaaS starter, this project should demonstrate security best practices. Currently no security headers are set on responses.

## Tasks

- [ ] Add Starlette middleware to set security headers on all responses:
  - `Strict-Transport-Security: max-age=31536000; includeSubDomains` (HSTS)
  - `X-Content-Type-Options: nosniff`
  - `X-Frame-Options: DENY`
  - `Content-Security-Policy: default-src 'self'` (configurable)
  - `Referrer-Policy: strict-origin-when-cross-origin`
  - `X-XSS-Protection: 0` (disabled in favor of CSP)
- [ ] Make headers configurable via environment variables (dev vs. prod)
- [ ] Add unit tests for header presence
- [ ] Document headers in README

## Success Criteria

- All security headers present on API responses
- Configurable for development (relaxed CSP) vs. production (strict)
- Tests verify header presence

jamescrowley321:
## Summary
- Add SecurityHeadersMiddleware with HSTS, X-Frame-Options, CSP, X-Content-Type-Options, Referrer-Policy
- Environment-aware: production enables HSTS + strict CSP; development relaxes CSP for localhost
- Configurable via ENVIRONMENT and CSP_POLICY env vars
- Outermost middleware — headers set on all responses including auth rejections

## Test plan
- [x] 147 unit tests pass
- [x] Security headers present on health endpoint
- [x] No HSTS in development mode
- [x] CSP includes localhost in development
- [x] Headers present on 401 responses

Automated PR from ralph loop task T64. Closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,21 @@ Four roles are defined via Terraform (`infra/rbac.tf`):
 
 Backend endpoints enforce authorization via `require_role()` and `require_permission()` dependency factories. Frontend uses `<RequireRole>` and `<RequirePermission>` components for conditional UI rendering.
 
+### Security Headers
+
+All API responses include security headers via `SecurityHeadersMiddleware`:
+
+| Header | Value |
+|--------|-------|
+| `X-Content-Type-Options` | `nosniff` |
+| `X-Frame-Options` | `DENY` |
+| `Referrer-Policy` | `strict-origin-when-cross-origin` |
+| `X-XSS-Protection` | `0` (disabled in favor of CSP) |
+| `Content-Security-Policy` | `default-src 'self'` (configurable via `CSP_POLICY`) |
+| `Strict-Transport-Security` | `max-age=31536000; includeSubDomains` (production only) |
+
+Set `ENVIRONMENT=production` to enable HSTS and strict CSP. Override CSP with `CSP_POLICY` env var.
+
 ## Project Structure
 
 ```


jamescrowley321:
## Adversarial Code Review -- PR #37

This PR is titled "feat: add security headers middleware" but actually delivers a much larger scope: security headers, RBAC, tenant management, user/member management, access keys, profile/attributes, database models, Descope Management API client, Terraform RBAC definitions, and full frontend pages. The security headers middleware itself is solid; the concerns below focus on the broader changes bundled in.

---

### MUST FIX

**1. Member management endpoints lack tenant-scoping on the target user (users.py)**
`deactivate_member`, `activate_member`, and `remove_member` accept an arbitrary `user_id` path parameter and call the Descope Management API directly *without verifying the target user belongs to the caller's tenant*. The `_tenant_id` dependency is resolved but never used for authorization -- it only proves the caller has a tenant, not that the target is in it. An admin in Tenant A can disable/delete a user who only belongs to Tenant B.

**File:** `backend/app/routers/users.py` lines 33-66

Compare with `accesskeys.py` which correctly calls `_verify_key_tenant()` before every mutation. The same pattern is needed here.

**2. `create_tenant` endpoint has no authorization check (tenants.py)**
Any authenticated user (even a viewer with no tenant context) can call `POST /api/tenants` to create unlimited tenants in your Descope project via the Management API. There is no role check or rate limit on this endpoint. This is a Management API operation that should be restricted to admins or owners.

**File:** `backend/app/routers/tenants.py` line 24-31

**3. `get_descope_client()` creates a new `httpx.AsyncClient` per API call (descope.py)**
Every method in `DescopeManagementClient` creates `async with httpx.AsyncClient() as client:` which opens and tears down a new connection pool on every single request. Under load this will exhaust file descriptors and add significant latency. The client should be shared (connection pooling) rather than created per-call.

**File:** `backend/app/services/descope.py` -- all methods

**4. `update_tenant_custom_attributes` has no allowlist (attributes.py)**
`PATCH /api/tenants/current/settings` accepts an arbitrary `dict[str, AttributeValue]` body and passes it directly to the Descope Management API. Unlike user profile updates (which enforce `ALLOWED_USER_ATTRIBUTES`), tenant custom attributes have no validation -- an admin can set arbitrary keys like `plan_tier` to escalate their own plan.

**File:** `backend/app/routers/attributes.py` lines 50-59

---

### SHOULD FIX

**5. CSP evaluated once at startup, `CSP_POLICY` env var is read at import time (security.py)**
The `CSP_POLICY` env var is read in `__init__` via `os.getenv()`. This is fine for normal operation, but the value is locked at middleware instantiation time. If the env var is changed (e.g., in a container orchestrator), a restart is required. This is a minor operational concern -- just worth documenting.

**6. No input validation on `invite_user` email (users.py)**
`InviteUserRequest` uses `email: str` with no email format validation. A malformed email will only fail at the Descope API level, returning an opaque error. Use Pydantic's `EmailStr` for immediate client-side feedback.

**File:** `backend/app/routers/users.py` line 11

**7. Bare `except Exception` swallows errors in profile/tenant endpoints (attributes.py, tenants.py)**
`get_profile`, `get_tenant_settings`, and `get_current_tenant` catch all exceptions and return empty/null data. This silently masks Descope API issues (auth failures, rate limits, misconfigurations) making debugging very difficult. At minimum, log the exception.

**Files:** `backend/app/routers/attributes.py` lines 30, 50; `backend/app/routers/tenants.py` line 57

**8. `_verify_key_tenant` calls the Descope API twice for mutations (accesskeys.py)**
For `deactivate`, `activate`, and `delete`, `_verify_key_tenant` loads the key (one API call), then the endpoint calls the mutation (second API call). The verify step also creates its *own* `get_descope_client()` instance rather than reusing the one from the endpoint. Consider passing the client in or combining the operations.

**File:** `backend/app/routers/accesskeys.py` line 17

**9. TOCTOU race in `_verify_key_tenant` (accesskeys.py)**
There is a time-of-check-to-time-of-use gap: after `_verify_key_tenant` confirms the key belongs to the tenant, the key could be reassigned before the subsequent `deactivate/activate/delete` call executes. Low probability but worth noting.

**10. Frontend invite allows selecting `owner` role without restriction (MemberManagement.tsx)**
The `AVAILABLE_ROLES` dropdown includes `owner`, but the backend only restricts owner assignment in the `/roles/assign` endpoint, not in `/members/invite`. An admin could invite a new user directly as `owner`, bypassing the escalation check.

**File:** `frontend/src/pages/MemberManagement.tsx` line 5; `backend/app/routers/users.py` lines 21-30

**11. Frontend `jwtDecode` does not handle non-standard base64 padding (jwt.ts)**
`atob()` may fail on JWTs with missing padding characters. While most JWT libraries produce valid base64url, adding padding normalization (`while (payload.length % 4) payload += '='`) before `atob()` would be more robust.

**File:** `frontend/src/utils/jwt.ts` line 4-5

---

### NITPICK

**12. PR title/scope mismatch**
The PR is titled "feat: add security headers middleware" but delivers ~3,800 lines across 30+ files covering RBAC, tenant management, member management, access keys, profiles, database models, Terraform config, and multiple frontend pages. This makes the PR very difficult to review and bisect. Consider splitting into focused PRs.

**13. `get_descope_client()` is not a FastAPI dependency (descope.py)**
It is called directly inside route handlers rather than injected via `Depends()`, which makes testing require `@patch` on the import path. Converting to a proper dependency would simplify mocking.

**14. `TenantResource.created_at` is stored as `str` (tenant.py)**
Using a string ISO format for timestamps works but loses the ability to do database-level date filtering/ordering. Consider `datetime` with a proper column type.

**File:** `backend/app/models/tenant.py` line 16

**15. Hardcoded role names across multiple files**
The strings `"owner"`, `"admin"`, `"member"`, `"viewer"` are duplicated across routers, tests, and frontend. A shared constants module would reduce drift risk.

**16. No pagination on list endpoints**
`list_access_keys`, `list_members`, `list_tenant_resources`, and `search_tenant_users` return all results without pagination. This will become a problem as data grows.

---

### Summary

The security headers middleware (the stated scope of the PR) is well-implemented. The main concerns are: (1) tenant isolation gaps in member management endpoints, (2) missing authorization on tenant creation, (3) httpx client lifecycle causing connection churn, and (4) unrestricted tenant custom attribute updates. Items 1 and 2 are the highest priority as they represent authorization bypass risks.

Issue #4, opened 2026-03-10
Phase 2a: Tenant Management


jamescrowley321:
## Description
Implement B2B multi-tenancy using Descope tenants.

**Descope features:** Tenants, tenant attributes, multi-tenant user assignment

## Tasks
- [ ] Org creation flow (new tenant provisioning via Descope Management API)
- [ ] Tenant switcher component (users can belong to multiple tenants)
- [ ] Tenant-scoped views (data isolated per tenant)
- [ ] Extract tenant from JWT claims (`dct` — Descope current tenant)
- [ ] Tenant-scoped database queries (SQLModel with tenant_id filtering)
- [ ] Tenant provisioning API endpoint
- [ ] Define default tenants and tenant attributes in Terraform

## Acceptance Criteria
- Users can create new organizations (tenants)
- Users belonging to multiple tenants can switch between them
- API responses are scoped to the current tenant
- No cross-tenant data leakage

## Dependencies
- #2, #3

## Phase
Phase 2: Multi-Tenancy & RBAC

Issue #3, opened 2026-03-10
Phase 1c: Session Management


jamescrowley321:
## Description
Implement proper session lifecycle management.

**Descope features:** Refresh token rotation, session token configuration

## Tasks
- [ ] Configure refresh token rotation via Terraform
- [ ] Session token validation on every API call (middleware)
- [ ] Logout flow — frontend clears session + backend-aware
- [ ] Session expiration handling in React (redirect to login)
- [ ] Configure token expiration settings via Terraform (session, refresh, inactivity)
- [ ] Handle token refresh transparently in frontend API client

## Acceptance Criteria
- Tokens refresh automatically before expiration
- Logout clears all session state
- Expired sessions redirect to login gracefully
- Inactivity timeout works as configured

## Dependencies
- #2

## Phase
Phase 1: Foundation & Core Auth

PR #58, opened 2026-03-25
Add auth audit logging for sensitive operations

### LINKED ISSUE #31: Add auth audit logging for sensitive operations
## Goal

Log all security-relevant operations for audit trail and compliance.

## Context

SaaS applications handling authentication must maintain audit logs of sensitive operations. This is a baseline requirement for SOC 2, ISO 27001, and similar compliance frameworks.

## Tasks

- [ ] Define audit event types: user_login, user_logout, role_change, permission_change, access_key_created, access_key_revoked, tenant_created, user_invited, user_removed
- [ ] Create audit log middleware/decorator that captures: timestamp, actor (user ID), action, target resource, IP address, result (success/failure)
- [ ] Store audit events (initially to structured log; future: database table or external service)
- [ ] Add audit log query endpoint for admin users
- [ ] Ensure audit logs cannot be tampered with (append-only pattern)
- [ ] Add tests for audit event generation

## Success Criteria

- All sensitive operations generate audit events
- Audit events include who, what, when, where, result
- Queryable audit trail for admins
- Tests verify audit events are generated correctly

jamescrowley321:
## Summary

- Add structured audit logging service (`app/services/audit.py`) with `AuditEventType` enum, `AuditEvent` Pydantic model, and convenience functions
- Instrument 15 security-sensitive endpoints with audit events: logout, role assign/remove, access key CRUD, user invite/deactivate/activate/remove, tenant create, profile update, tenant settings update
- Each audit event captures: timestamp, actor_id, tenant_id, target, IP address (with X-Forwarded-For support), result, and correlation ID via log context
- Replace ad-hoc access key logging with formal audit events
- Add 24 unit tests covering the audit module, IP extraction, and route integration
- Update README with audit logging documentation

## Test plan

- [x] All 194 unit tests pass
- [x] Audit event model creates with correct defaults and serializes all fields
- [x] `emit_audit_event` writes to dedicated `audit` logger with structured extra data
- [x] IP extraction handles X-Forwarded-For, direct client, and missing client
- [x] Route integration tests verify audit events emitted for all 15 endpoints
- [x] Content verification test confirms actor_id, tenant_id, target, result fields

Automated PR from ralph loop task T66. Closes #31

### DIFF
--- a/backend/app/routers/accesskeys.py
+++ b/backend/app/routers/accesskeys.py
@@ -36,13 +36,18 @@ async def create_access_key(
 ):
     """Create an access key scoped to the current tenant. Returns cleartext (shown once)."""
     client = get_descope_client()
-    result = await client.create_access_key(
-        name=body.name,
-        tenant_id=tenant_id,
-        expire_time=body.expire_time,
-        role_names=body.role_names,
-    )
-    audit_event(request, AuditEventType.ACCESS_KEY_CREATED, {"key_name": body.name})
+    target = {"key_name": body.name}
+    try:
+        result = await client.create_access_key(
+            name=body.name,
+            tenant_id=tenant_id,
+            expire_time=body.expire_time,
+            role_names=body.role_names,
+        )
+    except Exception:
+        audit_event(request, AuditEventType.ACCESS_KEY_CREATED, target, result="failure")
+        raise
+    audit_event(request, AuditEventType.ACCESS_KEY_CREATED, target)
     return result
 
 
@@ -78,8 +83,13 @@ async def deactivate_access_key(
     """Deactivate (revoke) an access key. Verifies key belongs to current tenant."""
     await _verify_key_tenant(key_id, tenant_id)
     client = get_descope_client()
-    await client.deactivate_access_key(key_id)
-    audit_event(request, AuditEventType.ACCESS_KEY_DEACTIVATED, {"key_id": key_id})
+    target = {"key_id": key_id}
+    try:
+        await client.deactivate_access_key(key_id)
+    except Exception:
+        audit_event(request, AuditEventType.ACCESS_KEY_DEACTIVATED, target, result="failure")
+        raise
+    audit_event(request, AuditEventType.ACCESS_KEY_DEACTIVATED, target)
     return {"status": "deactivated", "key_id": key_id}
 
 
@@ -93,8 +103,13 @@ async def activate_access_key(
     """Reactivate a previously deactivated access key. Verifies key belongs to current tenant."""
     await _verify_key_tenant(key_id, tenant_id)
     client = get_descope_client()
-    await client.activate_access_key(key_id)
-    audit_event(request, AuditEventType.ACCESS_KEY_ACTIVATED, {"key_id": key_id})
+    target = {"key_id": key_id}
+    try:
+        await client.activate_access_key(key_id)
+    except Exception:
+        audit_event(request, AuditEventType.ACCESS_KEY_ACTIVATED, target, result="failure")
+        raise
+    audit_event(request, AuditEventType.ACCESS_KEY_ACTIVATED, target)
     return {"status": "activated", "key_id": key_id}
 
 
@@ -108,6 +123,11 @@ async def delete_access_key(
     """Permanently delete an access key. Verifies key belongs to current tenant."""
     await _verify_key_tenant(key_id, tenant_id)
     client = get_descope_client()
-    await client.delete_access_key(key_id)
-    audit_event(request, AuditEventType.ACCESS_KEY_DELETED, {"key_id": key_id})
+    target = {"key_id": key_id}
+    try:
+        await client.delete_access_key(key_id)
+    except Exception:
+        audit_event(request, AuditEventType.ACCESS_KEY_DELETED, target, result="failure")
+        raise
+    audit_event(request, AuditEventType.ACCESS_KEY_DELETED, target)
     return {"status": "deleted", "key_id": key_id}
--- a/backend/app/routers/attributes.py
+++ b/backend/app/routers/attributes.py
@@ -55,8 +55,13 @@ async def update_profile_attribute(
     if body.key not in ALLOWED_USER_ATTRIBUTES:
         raise HTTPException(status_code=400, detail=f"Attribute '{body.key}' is not allowed")
     client = get_descope_client()
-    await client.update_user_custom_attribute(user_id, body.key, body.value)
-    audit_event(request, AuditEventType.PROFILE_UPDATED, {"key": body.key})
+    target = {"key": body.key}
+    try:
+        await client.update_user_custom_attribute(user_id, body.key, body.value)
+    except Exception:
+        audit_event(request, AuditEventType.PROFILE_UPDATED, target, result="failure")
+        raise
+    audit_event(request, AuditEventType.PROFILE_UPDATED, target)
     return {"status": "updated", "key": body.key, "value": body.value}
 
 
@@ -84,6 +89,11 @@ async def update_tenant_settings(
 ):
     """Update custom attributes on the current tenant. Requires owner or admin role."""
     client = get_descope_client()
-    await client.update_tenant_custom_attributes(tenant_id, body.custom_attributes)
-    audit_event(request, AuditEventType.TENANT_SETTINGS_UPDATED, {"tenant_id": tenant_id})
+    target = {"tenant_id": tenant_id}
+    try:
+        await client.update_tenant_custom_attributes(tenant_id, body.custom_attributes)
+    except Exception:
+        audit_event(request, AuditEventType.TENANT_SETTINGS_UPDATED, target, result="failure")
+        raise
+    audit_event(request, AuditEventType.TENANT_SETTINGS_UPDATED, target)
     return {"status": "updated", "tenant_id": tenant_id, "custom_attributes": body.custom_attributes}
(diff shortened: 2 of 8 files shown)


jamescrowley321:
## Adversarial Code Review -- PR #58

### MUST FIX

**1. Missing tenant membership verification on member management endpoints (Privilege Escalation / IDOR)**
`backend/app/routers/users.py` -- `deactivate_member`, `activate_member`, and `remove_member` accept an arbitrary `user_id` path parameter and use `get_tenant_id` + `require_role("owner", "admin")` but never verify the target `user_id` actually belongs to the caller's tenant. An admin in tenant A can deactivate/delete any user in tenant B by supplying their user_id. The `_tenant_id` dependency is resolved but never used to scope the Descope API call.

Compare with the access keys router, which correctly calls `_verify_key_tenant()` before operating. The users router needs an equivalent check -- load the user, confirm their tenant membership matches `_tenant_id`, then proceed.

**2. `POST /api/tenants` has no authorization check (Unauthorized Tenant Creation)**
`backend/app/routers/tenants.py` line `create_tenant` -- any authenticated user (even a viewer) can create new tenants via the Descope Management API. This endpoint only requires `get_claims` (authentication), not any role. The PR description says "Tenant isolation is enforced" but tenant *creation* is wide open. Should require owner/admin or be scoped to some privileged role.

**3. Audit event emitted after action but not on failure (Incomplete Audit Trail)**
Throughout all audited routes (e.g., `accesskeys.py`, `users.py`, `roles.py`), `audit_event()` is called only in the success path, after the Descope API call. If the Descope call raises an `httpx.HTTPStatusError`, the route propagates the exception as a 500 and no audit event is recorded. For compliance (SOC 2, ISO 27001 as claimed in the README), failed *attempts* are arguably more important to log than successes. Wrap in try/except and emit with `result="failure"` -- the `AuditEvent` model already supports this.

**4. X-Forwarded-For IP spoofing in audit logs**
`backend/app/services/audit.py` `_get_client_ip()` blindly trusts `X-Forwarded-For`. An attacker can set `X-Forwarded-For: 127.0.0.1` to forge their source IP in all audit records, undermining the forensic value of the audit log. This header should only be trusted if the app is behind a known reverse proxy. Consider either documenting this requirement prominently or adding a `TRUSTED_PROXIES` config to validate the immediate client before trusting forwarded headers.

---

### SHOULD FIX

**5. `DescopeManagementClient` instantiated per request (Performance)**
`backend/app/services/descope.py` -- `get_descope_client()` creates a new `DescopeManagementClient` on every call, and each method inside creates a new `httpx.AsyncClient()` (and therefore a new connection pool). This means every API call opens and tears down a TCP+TLS connection. Use a singleton client or at minimum a shared `httpx.AsyncClient` instance to benefit from connection pooling.

**6. `update_user_status` accepts arbitrary string -- no validation**
`backend/app/services/descope.py` `update_user_status()` takes `status: str` with no validation. The callers pass "enabled"/"disabled" but nothing prevents passing an arbitrary string to the Descope API. Use a `Literal["enabled", "disabled"]` type or an enum.

**7. Broad `except Exception` swallowing real errors**
`backend/app/routers/attributes.py` -- `get_profile()` and `get_tenant_settings()` catch bare `Exception` and return empty data with 200 OK. This hides Descope API errors (auth failures, rate limits, misconfiguration) from the caller and makes debugging difficult. At minimum, log the exception. Ideally, distinguish between "not found" (return empty) and "server error" (return 500 or 502).

**8. `InviteUserRequest.email` has no email format validation**
`backend/app/routers/users.py` -- The `email` field is a plain `str`. Pydantic's `EmailStr` type (from `pydantic[email]`) would prevent malformed/malicious inputs from reaching the Descope API.

**9. `CreateAccessKeyRequest.name` has no length/content validation**
`backend/app/routers/accesskeys.py` -- `name: str` allows empty strings, extremely long strings, or strings with control characters. Add `min_length=1, max_length=128` or similar constraints.

**10. `CreateTenantRequest.name` has no length validation**
Same as above -- `name: str` could be empty or excessively long.

**11. `RoleAssignmentRequest.role_names` allows arbitrary role names**
`backend/app/routers/roles.py` -- The list accepts any strings. An admin could attempt to assign non-existent roles. Validate against the known set `{"owner", "admin", "member", "viewer"}` to fail fast with a clear error.

**12. Rate limiting key function falls back to IP before auth middleware runs**
`backend/app/middleware/rate_limit.py` -- `get_rate_limit_key()` checks `request.state.claims` but the rate limiting middleware (`SlowAPIMiddleware`) is added *after* (and thus runs *before*) `TokenValidationMiddleware` in the middleware stack. So `request.state.claims` will typically not exist yet when the rate limiter resolves the key. Result: all rate limiting is effectively per-IP, never per-user-sub. The middleware order comment in `main.py` says "Rate limiting -- before auth so brute-force attempts are caught" which explains the intent, but the per-user keying described in the README/docs is misleading.

**13. `logout` in auth.py no longer imports `audit_event`**
Looking at the diff, the PR adds `from app.services.audit import AuditEventType, audit_event` and calls `audit_event(request, AuditEventType.USER_LOGOUT, ...)` in the logout route. But the current file on the base branch already has the rate limiter import without the audit import. Verify the audit import landed correctly on the feature branch -- if it was dropped during a rebase, the logout audit event call will raise a `NameError` at runtime.

**14. No error handling in frontend action handlers**
`frontend/src/pages/AccessKeys.tsx` -- `handleDeactivate`, `handleActivate`, `handleDelete` all `await apiFetch(...)` without try/catch and don't check `res.ok`. If the API call fails (403, 500), the error is silently swallowed and `loadKeys()` runs anyway, potentially confusing the user. Same pattern in `MemberManagement.tsx` for `handleToggleStatus` and `handleRemove`.

---

### NITPICK

**15. `_verify_key_tenant` is a module-level async function, not a class method**
`backend/app/routers/accesskeys.py` -- This works, but calling `get_descope_client()` inside creates a separate client from the one used in the route handler. Consider passing the client in, or making it a dependency.

**16. `DESCOPE_BASE_URL` duplicated**
`auth.py` defines `DESCOPE_BASE_URL = os.getenv(...)` at module level, while `descope.py` reads the same env var in `get_descope_client()`. Consolidate to avoid drift.

**17. `useApiClient` dependency array in `usePlanTier`**
`frontend/src/hooks/usePlanTier.ts` -- `apiFetch` is in the `useEffect` dependency array. If `useApiClient` returns a new function reference on each render (non-memoized), this will cause infinite re-fetches. Verify `apiFetch` is stable (wrapped in `useCallback`).

**18. `hasFetchedDashboard` / `hasFetchedIdToken` refs in Dashboard**
`frontend/src/pages/Dashboard.tsx` -- Using refs to prevent duplicate fetches suggests the dependency arrays may be overly broad. This pattern is fragile across StrictMode double-renders. Consider a proper data-fetching solution or `useEffect` cleanup.

**19. Audit `USER_LOGIN` enum member is defined but never used**
`backend/app/services/audit.py` -- `AuditEventType.USER_LOGIN = "user_login"` is defined but no endpoint emits it. Either remove it or add login audit events.

---

### Acceptance Audit

The PR description claims:
- Structured audit logging service -- **Implemented**
- 15 security-sensitive endpoints instrumented -- **Mostly complete**, but failure paths are not audited (see MUST FIX #3)
- IP address extraction with X-Forwarded-For -- **Implemented but spoofable** (see MUST FIX #4)
- 24 unit tests -- **Present and comprehensive for success paths**; no tests for audit-on-failure
- Replace ad-hoc access key logging -- **Verified**

The PR also includes substantial non-audit work (full RBAC system, tenant management, member management, access keys, frontend pages, Terraform RBAC/tenant definitions, security headers, rate limiting, structured logging) that is not mentioned in the PR title/description. This should either be split into separate PRs or the description should be updated to reflect the actual scope.

---

*Reviewed against commit on `feat/audit-logging` branch.*

jamescrowley321:
## Adversarial Code Review — PR #58 (Re-run 2026-03-27)

Reviewed against scoped diff: `feat/structured-logging` → `feat/audit-logging`

### MUST FIX

**[M1] X-Forwarded-For blindly trusted — IP spoofing in audit logs**
- File: `backend/app/services/audit.py:71-74`
- Issue: Any client can forge `X-Forwarded-For` to fabricate IP in audit events. For SOC 2/ISO 27001 compliance, this is an integrity flaw.
- Fix: Only trust `X-Forwarded-For` from known proxy CIDRs, or use `request.client.host` with proper proxy config.

**[M2] Audit events only log success — failed operations invisible**
- File: All route files (audit_event after Descope call, no try/except)
- Issue: If Descope raises, audit event never emitted. Failed security-sensitive operations go unlogged.
- Fix: Wrap Descope calls in try/except, emit `result="failure"` events before re-raising.

**[M3] PII (email) logged in audit events**
- File: `backend/app/routers/users.py:39`
- Issue: `user_invited` event logs invitee email. Under GDPR/CCPA, email is PII in logs with different retention/access controls.
- Fix: Log hashed/masked email or resulting `user_id` only.

### SHOULD FIX

**[S1]** Audit logging can crash request if logger misconfigured — wrap in try/except
**[S2]** Logout audit event fires even when logout was a no-op
**[S3]** No validation or sanitization of `target` dict contents
**[S4]** `create_tenant` lacks authorization check — any authenticated user (pre-existing, now audit-visible)
**[S5]** `user_id` in user management routes not validated against tenant membership (pre-existing IDOR)
**[S6]** `AuditEventType.USER_LOGIN` defined but never used
**[S7]** `audit_event()`/`emit_audit_event()` synchronous in async handlers — blocks event loop

### DEFER

[D1] No HTTP method/path in audit events — correlation ID links to request log
[D2] No audit for read operations — scoped to write operations in Phase 1
[D3] Timestamps as ISO strings — acceptable for log-only Phase 1
[D4] Tests don't verify `actor_id` matches mocked claims in most route tests

PR #185, opened 2026-04-01
feat: Error model, Result types, and RFC 9457 Problem Detail responses

jamescrowley321:
## Summary
- Add `IdentityError` base class with 6 domain-specific subclasses (`NotFound`, `Conflict`, `ValidationError`, `SyncFailed`, `ProviderError`, `Forbidden`)
- Add `ProblemDetailResponse` model implementing RFC 9457 with `application/problem+json` content type and `traceId` extension field
- Add `result_to_response()` helper that dispatches `expression.Result[T, IdentityError]` to JSON or Problem Detail responses
- Add comprehensive unit tests (355 lines) covering all error types, serialization, pattern matching, and edge cases

## Story
Refs #140
Part of PRD 5: Canonical Identity Domain Model

## Chained PR
Based on #184 — must be merged first

## Review Findings Addressed
- Security: 0 BLOCK, 1 WARN — no blocking issues
- Blind Hunter: 2 MUST FIX, 5 SHOULD FIX — all MUST FIX resolved (default branch, serialization guard, logging)
- Edge Cases: 2 paths found, 0 critical — non-blocking, mitigated with serialization guard
- Acceptance: 5 PASS, 0 FAIL, 0 PARTIAL
- Red Team: skipped (no auth/middleware files changed)

## Test plan
- [x] Unit tests pass (`make test-unit`)
- [x] Lint passes (`make lint`)
- [x] Independent review agents passed (blind, edge, acceptance, security)
- [ ] CI passes
- [ ] Manual verification against Descope sandbox

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #141, opened 2026-03-31
Epic 1 Story 1.4: OTel Instrumentation + Aspire Dashboard


jamescrowley321:
## Story 1.4: OTel Instrumentation + Aspire Dashboard

**Epic:** 1 — Canonical Identity Foundation
**FRs:** —
**Architecture:** D6, D12

As a system operator,
I want OpenTelemetry observability with trace visualization in Aspire Dashboard,
So that I can trace requests through the full write-through path and correlate errors to specific operations.

### Acceptance Criteria

**AC-1.4.1: Telemetry module**
**Given** `backend/app/telemetry.py`
**When** the module is initialized during app startup (before request handling)
**Then** OTel SDK is configured with `OTEL_SERVICE_NAME` (default: `identity-stack`)
**And** OTLP exporter sends to `OTEL_EXPORTER_OTLP_ENDPOINT` (default: `http://aspire-dashboard:18889`)
**And** FastAPI auto-instrumentor is registered (traces all HTTP routes)
**And** httpx auto-instrumentor is registered (traces Descope Management API calls)
**And** SQLAlchemy auto-instrumentor is registered with the async engine (traces Postgres queries)

**AC-1.4.2: CorrelationIdMiddleware removal**
**Given** the existing CorrelationIdMiddleware
**When** OTel is active
**Then** CorrelationIdMiddleware is removed from the middleware stack
**And** W3C `traceparent` header is used for distributed tracing
**And** existing tests that reference `X-Correlation-ID` are updated or removed

**AC-1.4.3: Error trace correlation**
**Given** RFC 9457 ProblemDetailResponse
**When** an error response is generated
**Then** the `traceId` field contains the current OTel trace ID

**AC-1.4.4: Graceful degradation**
**Given** OTel is disabled (no OTLP endpoint configured)
**When** the app starts
**Then** telemetry initialization is skipped gracefully (no crash)
**And** the app functions normally without OTel

### Technical Notes
- OTel packages: opentelemetry-sdk, opentelemetry-exporter-otlp-proto-grpc, opentelemetry-instrumentation-{fastapi,httpx,sqlalchemy,logging}
- Aspire Dashboard already in Docker Compose from Story 1.1
- Custom spans on IdentityService methods added in Epic 2

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #186, opened 2026-04-01
feat: OTel instrumentation with Aspire Dashboard support

jamescrowley321:
## Summary
- Add OpenTelemetry instrumentation module (`backend/app/telemetry.py`) with auto-instrumentors for FastAPI, HTTPX, SQLAlchemy, and logging
- Replace CorrelationIdMiddleware with OTel trace context (trace ID / span ID in log records)
- Add OTel dependencies to `pyproject.toml`
- Configure `OTEL_EXPORTER_OTLP_ENDPOINT` in Docker Compose for Aspire Dashboard integration
- Graceful degradation: telemetry is a no-op when endpoint is unset

## Story
Refs #141

## Chained PR
Based on #185 — must be merged first

## Review Findings Addressed
- Security: 0 BLOCK, 2 WARN, 4 INFO
- Blind Hunter: 3 MUST FIX (all resolved), 5 SHOULD FIX, 2 NITPICK
- Edge Cases: 3 paths found, 0 critical (all resolved — async_request_hook, json-logger version, double-init)
- Acceptance: 4 PASS, 0 FAIL, 0 PARTIAL
- Red Team: 0 critical, 0 high, 2 medium, 2 low — PASS

## Test plan
- [x] E2E tests pass (`make test-e2e`)
- [x] Independent review agents passed (blind, edge, acceptance, security, red team)
- [ ] CI passes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,3 +172,4 @@ jobs:
         run: docker compose build
         env:
           DESCOPE_PROJECT_ID: "test-project-id"
+          POSTGRES_PASSWORD: "ci-build-only"


jamescrowley321:

No BLOCK findings.

### WARN (should fix)

- [LIKELY] `backend/app/models/database.py:6` — Hardcoded default database credentials in application code
  The default `DATABASE_URL` fallback contains `identity:dev` credentials: `postgresql+asyncpg://identity:dev@localhost:5432/identity`. While these only apply when `DATABASE_URL` is unset and the password "dev" is a local-only default, this pattern means that if the app is deployed without `DATABASE_URL` explicitly set, it will attempt to connect with known credentials. The `docker-compose.yml` also uses `${POSTGRES_PASSWORD:-dev}` which defaults to the same "dev" password. In production, if `POSTGRES_PASSWORD` is not set in the environment, the database will be created with a well-known password.
  Mitigation: In `database.py`, fail loudly (raise `RuntimeError`) when `DATABASE_URL` is not set, similar to how `migrations/env.py` already does. The docker-compose default of "dev" is acceptable for local development but add a comment or startup check for production.

### INFO (acceptable risk)

- `docker-compose.yml:49` — Aspire Dashboard OTLP endpoint uses `DASHBOARD__OTLP__AUTHMODE: Unsecured`. This is acceptable for a local development dashboard. All service ports (postgres:5432, redis:6379, aspire:18888/4317) are bound to `127.0.0.1`, which is correct and prevents external access.

- `backend/app/telemetry.py:24` — The OTLP endpoint defaults to `http://aspire-dashboard:18889` (plaintext gRPC) when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset. This is internal service-to-service traffic within the Docker network, so plaintext is acceptable. The `use_insecure` flag (line 46) correctly switches to secure transport for `https://` endpoints.

- `backend/app/errors/problem_detail.py:119` — The `detail` field in RFC 9457 responses passes through `err.message` from `IdentityError`. This is acceptable because: (1) `IdentityError` messages are developer-controlled domain errors, not raw exception messages or stack traces, (2) no `IdentityError` subclass exposes internal state like SQL queries or file paths, and (3) the `_internal_error` fallback for unexpected failures uses a generic message ("Response serialization failed" / "Unexpected result type") rather than leaking exception details.

- `backend/app/middleware/correlation.py` (deleted) — The correlation ID middleware was removed in favor of OpenTelemetry tracing. The old middleware validated incoming `X-Correlation-ID` headers against a regex to prevent log injection. OTel trace/span IDs are internally generated (hex-formatted integers), so this log injection vector is eliminated rather than just migrated.

- `backend/migrations/versions/002_canonical_identity_schema.py:41-43` — Raw `op.execute()` SQL for enum creation uses only hardcoded string literals (no interpolation), so there is no injection risk.

- `backend/app/models/identity/provider.py:35` — The `Provider` model stores `config_ref` (a reference to Infisical/vault) rather than actual credentials. The comment on line 35 confirms: "No credentials stored -- config_ref points to Infisical." This is the correct pattern.

### Summary
- BLOCK: 0 | WARN: 1 | INFO: 6
- Overall: PASS

Issue #144, opened 2026-03-31
Epic 2 Story 2.1: User Service + Descope Sync Adapter


jamescrowley321:
## Story 2.1: User Service + Descope Sync Adapter

**Epic:** 2 — Identity & Access Administration
**FRs:** FR1, FR3, FR4, FR23-25
**Architecture:** D4, D7, D8

As an admin,
I want user CRUD operations backed by canonical Postgres with automatic sync to Descope,
So that the canonical store owns user data while keeping Descope in sync.

### Acceptance Criteria

**AC-2.1.1: User CRUD**
**Given** `PostgresIdentityService.create_user()` is called
**Then** a User record is created in Postgres with `created_at` and `updated_at` timestamps
**And** `DescopeSyncAdapter.sync_user()` is called after the Postgres write
**And** if sync succeeds, `Ok(user)` is returned
**And** if sync fails, the Postgres write is NOT rolled back, a warning is logged (operation/payload/error/timestamp), and `Ok(user)` is still returned

**AC-2.1.2: User search**
**Given** `PostgresIdentityService.search_users()` is called with filter params (email, name, status, tenant_id)
**Then** results are filtered from Postgres using the provided criteria
**And** the query scopes by `tenant_id` when provided

**AC-2.1.3: User deactivation (soft delete)**
**Given** `PostgresIdentityService.deactivate_user()` is called
**Then** the user's status is set to `inactive` (soft delete, not hard delete)
**And** `updated_at` is updated
**And** sync to Descope is attempted

**AC-2.1.4: Descope sync adapter**
**Given** `DescopeSyncAdapter` in `backend/app/services/adapters/descope.py`
**Then** it uses the existing `DescopeManagementClient` for API calls
**And** each method wraps httpx calls returning `Result[None, SyncError]`
**And** each method has an OTel span (`descope.sync_user`, etc.)

**AC-2.1.5: Service method invariants**
**Given** every IdentityService method
**Then** it has an OTel span with `tenant.id` attribute
**And** it takes `tenant_id: UUID` as explicit parameter
**And** it returns `Result[T, IdentityError]`

### Technical Notes
- Write-through: Postgres first, sync second (D7). Sync failure → log, never rollback (NFR9)
- SyncFailed is logged at service level; router decides HTTP status

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #195, opened 2026-04-05
feat: User Service + Descope Sync Adapter

jamescrowley321:
## Summary
- Add `PostgresIdentityService` implementing user CRUD (create, get, update, deactivate, search) with write-through pattern: Postgres first, Descope sync second
- Add `DescopeSyncAdapter` wrapping `DescopeManagementClient` for user sync with OTel spans and Result-based error handling
- Wire up `get_identity_service()` dependency to return real service instance with `DescopeSyncAdapter`
- Sync failures are logged but never rollback the Postgres write (D7 write-through pattern)
- All service methods return `Result[T, IdentityError]`, use explicit `tenant_id` scoping, and include OTel spans

## Story
Refs #144

## Chained PR
Based on #194 — must be merged first

## Review Findings Addressed
- Security: 0 BLOCK, 3 WARN — no blocking issues
- Blind Hunter: 3 MUST FIX (1 fixed: status enum validation; 2 accepted: TOCTOU is D7 by-design, create_user lifecycle is documented), 5 SHOULD FIX
- Edge Cases: 2 unhandled paths found, 2 fixed (invalid status crash, update_user error message)
- Acceptance: 16 PASS, 0 FAIL, 0 PARTIAL
- Red Team: skipped (no auth-related files changed)

## Test plan
- [ ] CI passes

### DIFF
--- a/backend/app/services/identity_impl.py
+++ b/backend/app/services/identity_impl.py
@@ -154,7 +154,7 @@ class PostgresIdentityService(IdentityService):
                 await self._session.flush()
             except IntegrityError:
                 await self._session.rollback()
-                return Error(Conflict(message=f"User with email '{email}' already exists"))
+                return Error(Conflict(message=f"User with email '{user.email}' already exists"))
 
             user_dict = self._user_to_dict(user)
 
@@ -226,7 +226,11 @@ class PostgresIdentityService(IdentityService):
             )
 
             if status:
-                stmt = stmt.where(User.status == status)
+                try:
+                    status_enum = UserStatus(status)
+                except ValueError:
+                    return Ok([])  # Invalid status matches no users
+                stmt = stmt.where(User.status == status_enum)
 
             if query:
                 # Escape ILIKE metacharacters to prevent wildcard injection
--- a/backend/tests/unit/test_identity_impl.py
+++ b/backend/tests/unit/test_identity_impl.py
@@ -241,6 +241,23 @@ class TestUpdateUser:
         result = await service.update_user(tenant_id=tenant_id, user_id=user.id, email="taken@example.com")
         assert result.is_error()
         assert isinstance(result.error, Conflict)
+        # Error message should show the actual email on the user model, not the param
+        assert "taken@example.com" in result.error.message
+
+    @pytest.mark.anyio
+    async def test_integrity_error_message_uses_model_email(self, service, mock_session, tenant_id):
+        """When only user_name is updated but IntegrityError occurs, message uses user.email not None."""
+        user = _make_user(email="real@example.com")
+        exec_result = MagicMock()
+        exec_result.scalar_one_or_none.return_value = user
+        mock_session.execute = AsyncMock(return_value=exec_result)
+        mock_session.flush = AsyncMock(side_effect=IntegrityError("dup", {}, None))
+        mock_session.rollback = AsyncMock()
+
+        result = await service.update_user(tenant_id=tenant_id, user_id=user.id, user_name="newname")
+        assert result.is_error()
+        assert isinstance(result.error, Conflict)
+        assert "real@example.com" in result.error.message
 
     @pytest.mark.anyio
     async def test_partial_update_only_changes_given_fields(self, service, mock_session, tenant_id):
@@ -416,6 +433,13 @@ class TestSearchUsers:
         assert result.is_ok()
         assert len(result.ok) == 1
         assert result.ok[0]["status"] == "inactive"
+
+    @pytest.mark.anyio
+    async def test_invalid_status_returns_empty(self, service, mock_session, tenant_id):
+        """Invalid status string returns empty list instead of crashing."""
+        result = await service.search_users(tenant_id=tenant_id, status="bogus")
+        assert result.is_ok()
+        assert result.ok == []
 
 
 # ---------------------------------------------------------------------------


jamescrowley321:
# Edge Case Review

## Finding 1: `get_descope_client()` crashes the dependency factory when env vars are missing

- **Location**: `backend/app/dependencies/identity.py:19` / `backend/app/services/descope.py:395-396`
- **Trigger Condition**: The app starts without `DESCOPE_PROJECT_ID` or `DESCOPE_MANAGEMENT_KEY` in the environment, and the singleton was not initialized via `init_descope_client()` during lifespan (e.g., a unit test or integration test that does not mock `get_descope_client`, or a misconfigured deployment).
- **Guard Snippet**:
  ```python
  # In get_identity_service, catch and convert to a service-level error,
  # or ensure init_descope_client() is always called before routes are live.
  ```
- **Consequence**: [CRASH] `get_descope_client()` falls through to `os.environ["DESCOPE_PROJECT_ID"]` which raises `KeyError`. This unhandled exception propagates as an HTTP 500 on every request that injects `IdentityService`. The previous stub (`raise NotImplementedError`) was at least explicit; now a missing env var produces a cryptic `KeyError` in the dependency factory.

## Finding 2: `sync_user` calls `load_user` then `resolve_login_id` -- double network round-trip for the same data

- **Location**: `backend/app/services/adapters/descope.py:40-42`
- **Trigger Condition**: Every `sync_user` call for a user that exists in Descope.
- **Guard Snippet**:
  ```python
  # resolve_login_id already calls load_user internally (descope.py:128).
  # Replace lines 40-42 with just:
  login_id = await self._client.resolve_login_id(str(user_id))
  # and handle 404 from resolve_login_id instead.
  ```
- **Consequence**: [DEGRADED] `load_user` is called on line 40, then `resolve_login_id` on line 42 calls `load_user` again internally (descope.py line 128). Every successful `sync_user` makes 3 HTTP round-trips (load + load-via-resolve + update_status) instead of the necessary 2. Under load, this doubles Descope API quota consumption for the load endpoint.

## Finding 3: `resolve_login_id` 404 is not caught as a 404 in `sync_user` inner try-block

- **Location**: `backend/app/services/adapters/descope.py:40-52`
- **Trigger Condition**: `load_user` succeeds (user exists) but `resolve_login_id` internally calls `load_user` again and this second call returns 404 (race condition: user deleted between the two calls).
- **Guard Snippet**:
  ```python
  # The inner except block on line 44 only catches HTTPStatusError from the
  # explicit load_user call. If resolve_login_id raises 404, the except on
  # line 44 catches it and silently treats it as "user not found, skip sync"
  # -- but the user DID exist moments ago. This is benign but masks a race.
  ```
- **Consequence**: [WRONG] If a user is deleted in Descope between the `load_user` (line 40) and `resolve_login_id` (line 42), the 404 from `resolve_login_id` is caught by the inner `except httpx.HTTPStatusError` block (line 44) and treated as "user not found, skip sync" -- returning `Ok(None)`. The caller believes sync succeeded when the user was actually deleted mid-operation. The `update_user_status` call is silently skipped without logging.

## Finding 4: Session state is poisoned after `IntegrityError` rollback in `create_user`

- **Location**: `backend/app/services/identity_impl.py:81-83`
- **Trigger Condition**: `create_user` hits a duplicate email (`IntegrityError`), `rollback()` is called, and the caller's request handler continues to use the same session for additional operations within the same request scope.
- **Guard Snippet**:
  ```python
  # After session.rollback(), the session's identity map is cleared and
  # any previously loaded objects become detached. If the same request
  # later calls another service method using this session, those objects
  # will raise DetachedInstanceError on attribute access.
  # Consider using SAVEPOINT (begin_nested) instead of full rollback.
  ```
- **Consequence**: [CRASH] SQLAlchemy's `session.rollback()` invalidates the entire transaction. If any FastAPI route handler calls `create_user`, gets a `Conflict` error, then attempts to use the same injected session for another query (e.g., logging, audit, or a follow-up operation), the session is in an invalid state. Accessing attributes of any previously-loaded ORM objects will raise `sqlalchemy.orm.exc.DetachedInstanceError`. The same applies to `update_user` (line 155-156) and `deactivate_user` (line 194-195).

## Finding 5: `search_users` has no pagination -- unbounded result set

- **Location**: `backend/app/services/identity_impl.py:214-249`
- **Trigger Condition**: A tenant has thousands of users and `search_users` is called with no query filter (or a very broad one).
- **Guard Snippet**:
  ```python
  # Add limit/offset parameters:
  # stmt = stmt.limit(limit).offset(offset)
  ```
- **Consequence**: [DEGRADED] The query loads every matching `User` row into memory via `.scalars().all()`, serializes all of them into dicts, and returns a single list. For tenants with large user counts, this causes excessive memory consumption, slow response times, and potential OOM. There is no `LIMIT` clause.

## Finding 6: `update_user` accepts a no-op call with all None fields

- **Location**: `backend/app/services/identity_impl.py:120-173`
- **Trigger Condition**: `update_user` is called with `tenant_id` and `user_id` but all optional fields (`email`, `user_name`, `given_name`, `family_name`) are `None`.
- **Guard Snippet**:
  ```python
  if all(v is None for v in (email, user_name, given_name, family_name)):
      return Ok(self._user_to_dict(user))  # Skip flush and sync
  ```
- **Consequence**: [DEGRADED] The method still queries the user, sets `updated_at` to now, flushes to the database (triggering an UPDATE with only the timestamp change), and then fires a sync to the Descope API. This is a wasted write + network call. The `updated_at` timestamp changes even though no user-visible data was modified, which pollutes audit trails and triggers unnecessary IdP syncs.

## Finding 7: `DescopeSyncAdapter` status mapping treats any non-"inactive" status as "enabled"

- **Location**: `backend/app/services/adapters/descope.py:35-36`
- **Trigger Condition**: `data["status"]` is `"provisioned"` (a valid `UserStatus` enum value per user.py line 14).
- **Guard Snippet**:
  ```python
  # The UserStatus enum has three values: active, inactive, provisioned.
  # Currently: descope_status = "disabled" if status == "inactive" else "enabled"
  # A "provisioned" user (not yet active) would be synced as "enabled" in Descope.
  ```
- **Consequence**: [WRONG] A user with `UserStatus.provisioned` (a pending/not-yet-active state) is synced to Descope as `"enabled"`, granting them full access in the IdP before they have completed onboarding. The mapping should explicitly handle all three enum variants.

PR #203, opened 2026-04-06
feat: User Service + Descope Sync Adapter (onion architecture)

jamescrowley321:
## Summary
- Add `UserRepository` with CRUD operations and tenant-scoped search using SQLAlchemy (inner layer)
- Add `UserService` domain orchestration with Result-based returns, OTel spans, and tenant_id scoping (middle layer)
- Add `DescopeSyncAdapter` wrapping DescopeManagementClient with fire-and-forget sync pattern (outer layer)
- Add `get_user_service()` DI factory wiring all onion layers via FastAPI dependency injection
- Onion architecture: repositories (DB) → domain services (logic) → adapters (sync) — no layer depends on an outer layer

## Story
Refs #144

## Chained PR
First PR in Epic 2 — based on main, no prior PR dependency.

## Review Findings Addressed
- Security: 0 BLOCK, 3 WARN — all deferred by design to stories 2.2/2.3
- Blind Hunter: 0 MUST FIX, findings resolved in review-fix
- Edge Cases: 0 CRASH within scope, all paths resolved
- Acceptance: 7/7 PASS — all ACs satisfied

### Fixed in review-fix (2 commits):
1. ILIKE wildcard escape in repository search
2. sync_user error on missing email/status (was silent Ok)
3. Status mapping fail-closed via explicit allowlist
4. IntegrityError TOCTOU handling → RepositoryConflictError
5. AC-2.1.6: tenant_id param + tenant.id span on all methods
6. Commit-before-sync ordering (prevents IdP ghost state)
7. _log_sync_failure uses public Result API
8. Dead except blocks removed from stub methods
9. Generic conflict message for non-email constraint violations

## Test plan
- [x] Independent review agents passed (acceptance, blind-hunt, edge-case, security)
- [ ] CI passes

### DIFF
--- a/backend/app/services/user.py
+++ b/backend/app/services/user.py
@@ -132,7 +132,7 @@ class UserService:
             try:
                 user = await self._repository.update(user)
             except RepositoryConflictError:
-                return Error(Conflict(message=f"User with email '{email}' already exists"))
+                return Error(Conflict(message="User update conflicts with existing data"))
 
             result_dict = user.model_dump()
             sync_data = {"email": user.email, "status": user.status.value}
--- a/backend/tests/unit/test_user_service.py
+++ b/backend/tests/unit/test_user_service.py
@@ -241,6 +241,7 @@ class TestUpdateUser:
 
         assert result.is_error()
         assert isinstance(result.error, Conflict)
+        assert "conflicts" in result.error.message
 
 
 @pytest.mark.anyio


jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `services/user.py:44` | `create_user` accepts `tenant_id` (used as span attribute on line 55) but never creates a `UserTenantRole` row linking the user to that tenant. | After `repository.create(user)`, also insert a `UserTenantRole(user_id=user.id, tenant_id=tenant_id, role_id=...)` or call a separate assignment method. | [WRONG] Created user is invisible to `search_users` which JOINs through `user_tenant_roles` on the tenant_id. Every user created via this service is immediately orphaned from tenant-scoped queries. |
| `services/user.py:75` | `commit()` in `create_user` is unguarded. If `commit()` raises (connection dropped, lock timeout, serialization failure), the exception propagates as an unhandled 500 to the caller. | Wrap in `try: await self._repository.commit() except Exception as exc: return Error(IdentityError(message="Failed to persist user"))` or let a middleware translate DB exceptions. | [CRASH] |
| `services/user.py:139` | Same unguarded `commit()` in `update_user`. Any DB-level failure during commit propagates as a raw 500. | Same guard as above. | [CRASH] |
| `services/user.py:172` | Same unguarded `commit()` in `deactivate_user`. | Same guard as above. | [CRASH] |
| `services/user.py:132-135` | In `update_user`, when `email` is `None` (no email update requested) but a different field's flush triggers `RepositoryConflictError` (e.g., a future unique constraint on `user_name`), the error message is `f"User with email '{None}' already exists"` -- blaming email when it was not changed. | Use: `msg = f"Email '{email}' already exists" if email is not None else "User update conflict"` | [WRONG] |
| `services/user.py:167-168` | `deactivate_user` does not check if user is already inactive. Calling it on an already-inactive user fires a redundant `flush()`, `commit()`, and Descope `sync_user` call for no state change. The `updated_at` timestamp also changes spuriously. | Add early return: `if user.status == UserStatus.inactive: return Ok(user.model_dump())` | [DEGRADED] |
| `dependencies/identity.py:29` | If the singleton was never initialized via `init_descope_client()` and `DESCOPE_PROJECT_ID` or `DESCOPE_MANAGEMENT_KEY` env vars are missing, `get_descope_client()` raises a bare `KeyError` (from `os.environ[...]` dict access) which surfaces as a 500 with a raw traceback leaking the env var name. | In `get_descope_client` fallback path, catch `KeyError` and raise `RuntimeError("Descope client not initialized and DESCOPE_PROJECT_ID/DESCOPE_MANAGEMENT_KEY not set")`. | [CRASH] |
| `services/adapters/descope.py:49-51` | `sync_user` checks `if not email` which passes for whitespace-only strings like `" "`. A whitespace email is sent to Descope's `update_user_status` as the `loginId`, which will likely fail at the API level. The exception is caught and returned as `SyncError`, so it does not crash, but the sync silently fails with a misleading error message from Descope rather than a clear validation message. | Tighten check: `if not email or not email.strip() or not status:` | [DEGRADED] |
| `services/adapters/descope.py:91-94` | `sync_role` defaults `name` to `""` via `data.get("name", "")`. Unlike `sync_user` which validates required fields, `sync_role` sends an empty string name to Descope's `create_role` endpoint. The Descope API call will fail and be caught by the except block, but the error message will be an opaque API error rather than a clear validation failure. | Add validation like `sync_user` does: `if not name: return Error(SyncError(message="Missing required sync data: name", ...))` | [DEGRADED] |
| `services/adapters/descope.py:116-118` | Same issue in `sync_permission`: `name` defaults to `""` with no validation before sending to Descope. | Add name validation. | [DEGRADED] |
| `services/adapters/descope.py:140-142` | Same issue in `sync_tenant`: `name` defaults to `""` with no validation before sending to Descope. | Add name validation. | [DEGRADED] |

### Summary
- Unhandled paths found: 11
- Critical (crash/data loss): 4
- Non-critical (wrong result/degraded): 7

### Notes

**Most impactful finding -- missing tenant association (services/user.py:44):** `create_user` accepts `tenant_id` and sets it as an OTel span attribute, but never creates a `UserTenantRole` row. Since `search_users` (line 195) queries through a JOIN on `user_tenant_roles`, every user created via this service is immediately invisible to tenant-scoped search. This is an architectural gap, not a race condition -- it affects 100% of create calls.

**Unguarded commit pattern (3 instances):** All three mutating methods (`create_user`, `update_user`, `deactivate_user`) leave `commit()` unguarded. While `flush()` failures are caught via `RepositoryConflictError`, `commit()` can independently fail (network partition, serialization conflict in PostgreSQL's SERIALIZABLE isolation, lock wait timeout). These would surface as raw 500s.

**Corrected from previous review:** The previous version of this review reported three false positives: (1) `IntegrityError` TOCTOU races in create/update -- these ARE handled by the `RepositoryConflictError` catch blocks on lines 69 and 134; (2) LIKE wildcard injection in search -- the `_escape_like()` helper on line 23-25 of `repository/user.py` correctly escapes `%`, `_`, and `\`; (3) sync-before-commit ordering -- sync actually runs AFTER `commit()` in all three methods (lines 75-78, 139-142, 172-175).

jamescrowley321:
## Review: Acceptance Auditor

### PASS
- [AC-2.1.1] **User repository** — `UserRepository` in `backend/app/repositories/user.py:28-121` handles all SQLAlchemy queries for user CRUD: `create` (:37), `get` (:50), `get_by_email` (:54), `update` (:60), `search` (:72). Every tenant-scoped query method (`search`) takes `tenant_id` as explicit parameter (:75) and scopes via JOIN on `UserTenantRole` (:87). Contains no business logic, no OTel spans, no adapter calls — pure data access. Takes `AsyncSession` via constructor injection (:34).

- [AC-2.1.2] **User domain service** — `UserService` in `backend/app/services/user.py:25-214` orchestrates `UserRepository` and `IdentityProviderAdapter`. Zero SQLAlchemy imports (verified via grep). `create_user()` persists via `repository.create()` (:68), commits (:75), then calls `adapter.sync_user()` (:77-81). If sync fails, `_log_sync_failure()` (:201-214) logs a warning with operation and error; `Ok(result_dict)` is returned (:83) regardless. Postgres write is NOT rolled back on sync failure — commit at :75 precedes the sync call. Tested at `test_user_service.py`: `test_create_user_success` (:57), `test_create_user_commit_before_sync` (:77, verifies commit-before-sync ordering), `test_create_user_sync_failure_still_returns_ok` (:126).

- [AC-2.1.3] **User search** — `UserService.search_users()` at `user.py:182-199` delegates to `UserRepository.search()` with `tenant_id`. Repository `search()` (:72-112) filters by `tenant_id` via JOIN on `UserTenantRole` (:87), with optional filters for email (:92-94), name (:96-104), and status (:106-107). Tested at `test_user_service.py:287-319`: verifies delegation, tenant scoping, and query-to-name filter mapping.

- [AC-2.1.4] **User deactivation (soft delete)** — `UserService.deactivate_user()` at `user.py:149-180` sets `user.status = UserStatus.inactive` (:167), calls `repository.update()` (:168) — soft delete, not hard delete. `updated_at` is updated via SQLAlchemy `onupdate=sa.func.now()` on the User model (`models/identity/user.py:43`). Sync to Descope attempted via adapter (:174-178). Tested at `test_user_service.py:246-284`: `test_deactivate_user_success` (:250, asserts `user.status == UserStatus.inactive`), `test_deactivate_sync_failure_still_returns_ok` (:273).

- [AC-2.1.5] **Descope sync adapter** — `DescopeSyncAdapter` in `backend/app/services/adapters/descope.py:27-78` uses `DescopeManagementClient` (injected at :36). `sync_user()` (:39) wraps `client.update_user_status()` (:68) returning `Result[None, SyncError]`. OTel span named `descope.sync_user` (:46). Maps canonical statuses to Descope statuses via `_STATUS_MAP` (:20-24). Input validation returns `Error(SyncError)` for missing data (:51-58) or unknown status (:59-67). Client exceptions caught and wrapped as `Error(SyncError)` (:70-78). Tested at `test_descope_adapter.py:32-107`: covers active/inactive/provisioned mapping, missing data, unknown status, and client error paths.

- [AC-2.1.6] **Domain service method invariants** — All five `UserService` methods satisfy all four invariants:
  - **OTel spans with `tenant.id`**: `create_user` (:54-55), `get_user` (:92-93), `update_user` (:112-113), `deactivate_user` (:159-160), `search_users` (:192-193) — all set `tenant.id` span attribute.
  - **`tenant_id: UUID` as explicit parameter**: `create_user` (:44), `get_user` (:88), `update_user` (:104), `deactivate_user` (:152), `search_users` (:185) — all accept `tenant_id`.
  - **Returns `Result[T, IdentityError]`**: All methods annotated correctly — `Result[dict, IdentityError]` for single-entity methods, `Result[list[dict], IdentityError]` for search.
  - **Calls repository for data access, never SQLAlchemy directly**: Zero SQLAlchemy imports in `user.py`.

- [AC-2.1.7] **DI wiring** — `get_user_service()` in `backend/app/dependencies/identity.py:19-30` composes: `AsyncSession -> UserRepository(session)` (:28), `DescopeManagementClient -> DescopeSyncAdapter(client)` (:29), `-> UserService(repository=repository, adapter=adapter)` (:30). Repository depends on session (inner layer), service depends on repo + adapter (middle layer), no layer depends on an outer layer. Tested at `test_identity_dependency.py:17-53`: verifies `isinstance` checks for all components and that session/client are correctly threaded through.

### FAIL
_None._

### PARTIAL
_None._

### SCOPE CREEP
- **Adapter methods beyond User scope**: `DescopeSyncAdapter` implements `sync_role()` (:80-104), `sync_permission()` (:106-128), `sync_tenant()` (:130-152) with full client call logic, plus `sync_role_assignment()` (:154-169), `delete_user()` (:171-178), `delete_role()` (:180-184), `delete_permission()` (:186-190), `delete_tenant()` (:192-196) as placeholder stubs. These are all required by the `IdentityProviderAdapter` ABC contract, so the concrete adapter must implement them. The delete methods and `sync_role_assignment` are minimal stubs returning `Ok(None)`. The `sync_role/permission/tenant` methods have full implementations with client calls. Acceptable as forward scaffolding mandated by the ABC, but the full implementations (vs. stubs) go slightly beyond what Story 2.1 requires.
- **Tests for non-User adapter methods**: `test_descope_adapter.py:110-205` tests `sync_role`, `sync_permission`, `sync_tenant`, and all `delete_*` stubs. Beyond Story 2.1 scope but consistent with the implemented code.

### Architecture Violations
_None._ The onion architecture layering is correctly maintained:
- **Inner (data access)**: `UserRepository` depends only on SQLAlchemy/session — no imports from services or adapters.
- **Middle (domain logic)**: `UserService` depends on `UserRepository` and `IdentityProviderAdapter` (ABC) — no SQLAlchemy imports, no concrete adapter dependency.
- **Outer (infrastructure)**: `DescopeSyncAdapter` implements `IdentityProviderAdapter` and depends on `DescopeManagementClient`.
- **DI composition**: `get_user_service()` assembles layers at the composition root without violating dependency direction.

### Summary
- Total ACs: 7
- Pass: 7 | Fail: 0 | Partial: 0

Issue #145, opened 2026-03-31
Epic 2 Story 2.2: Role, Permission + Tenant Service


jamescrowley321:
## Story 2.2: Role, Permission + Tenant Service

**Epic:** 2 — Identity & Access Administration
**FRs:** FR5-12
**Architecture:** D4, D7, D8

As an admin,
I want role, permission, and tenant CRUD with role-permission mapping and user-role assignment backed by canonical Postgres,
So that all authorization configuration is owned by the canonical store.

### Acceptance Criteria

**AC-2.2.1: Role repository + service**
**Given** `RoleRepository` in `backend/app/repositories/role.py` and `RoleService` in `backend/app/services/role.py`
**Then** `RoleRepository` handles all SQLAlchemy queries for role CRUD and role-permission mappings
**And** `RoleService` orchestrates role CRUD by calling `RoleRepository` for persistence and adapter for sync
**And** `create_role()` called with name, description, optional tenant_id creates a Role in Postgres (tenant_id=NULL for global roles)
**And** sync to Descope via `DescopeSyncAdapter.sync_role()`
**And** `Result[Role, IdentityError]` is returned

**AC-2.2.2: Permission repository + service**
**Given** `PermissionRepository` in `backend/app/repositories/permission.py` and `PermissionService` in `backend/app/services/permission.py`
**Then** `PermissionRepository` handles all SQLAlchemy queries for permission CRUD
**And** `PermissionService.create_permission()` persists via repository and syncs to Descope
**When** `map_permission_to_role()` is called
**Then** a RolePermission mapping is created via repository and synced

**AC-2.2.3: Tenant repository + service**
**Given** `TenantRepository` in `backend/app/repositories/tenant.py` and `TenantService` in `backend/app/services/tenant.py`
**Then** `TenantRepository` handles all SQLAlchemy queries for tenant CRUD
**And** `TenantService.create_tenant()` called with name and domains creates a Tenant via repository and syncs
**When** `get_tenant_users_with_roles()` is called
**Then** the repository returns all users in the tenant with role assignments

**AC-2.2.4: User-tenant-role assignment**
**Given** `assign_role_to_user()` called with user_id, tenant_id, role_id, assigned_by
**Then** a UserTenantRole record is created via `UserTenantRoleRepository` with `assigned_by` and `assigned_at`
**And** sync to Descope is attempted via adapter
**When** the same assignment already exists
**Then** `Error(Conflict(...))` is returned

[...]

PR #204, opened 2026-04-06
feat: Role/Permission/Tenant repositories, services, and DI wiring

jamescrowley321:
## Summary
- Add `RoleRepository`, `PermissionRepository`, `TenantRepository`, and `UserTenantRoleRepository` (inner layer — SQLAlchemy only)
- Add `RoleService`, `PermissionService`, `TenantService` domain services (business logic, OTel spans, adapter sync)
- Wire DI factories (`get_role_service`, `get_permission_service`, `get_tenant_service`) in identity dependencies
- Implement `sync_role_assignment` in DescopeSyncAdapter (was placeholder)
- 44 new unit tests covering all services, repos, DI wiring, and adapter paths

## Story
Refs #145

## Chained PR
Based on #203 — merge first

## Review Findings Addressed
- Security: 0 BLOCK, 1 WARN — resolved (login_id fallback deferred to 2.3)
- Blind Hunter: 0 MUST FIX, 3 SHOULD FIX, residual defensive coding beyond story scope
- Edge Cases: 4 paths — 1 CRASH (rare commit failure), 3 DEGRADED — accepted residual
- Acceptance: 6/6 PASS

## Test plan
- [x] Unit tests pass (`make test-unit`) — 1003 tests
- [ ] CI passes

### DIFF
--- a/backend/app/services/permission.py
+++ b/backend/app/services/permission.py
@@ -61,6 +61,7 @@ class PermissionService:
             try:
                 permission = await self._repository.create(permission)
             except RepositoryConflictError:
+                await self._repository.rollback()
                 return Error(Conflict(message=f"Permission '{name}' already exists"))
 
             result_dict = permission.model_dump()
--- a/backend/app/services/role.py
+++ b/backend/app/services/role.py
@@ -73,6 +73,7 @@ class RoleService:
             try:
                 role = await self._repository.create(role)
             except RepositoryConflictError:
+                await self._repository.rollback()
                 return Error(Conflict(message=f"Role '{name}' already exists in this scope"))
 
             result_dict = role.model_dump()
@@ -129,6 +130,7 @@ class RoleService:
             try:
                 mapping = await self._repository.add_permission(role_id, permission_id)
             except RepositoryConflictError:
+                await self._repository.rollback()
                 return Error(Conflict(message=f"Permission '{permission_id}' is already mapped to role '{role_id}'"))
 
             result_dict = {
@@ -188,6 +190,7 @@ class RoleService:
             try:
                 assignment = await self._assignment_repository.create(assignment)
             except RepositoryConflictError:
+                await self._assignment_repository.rollback()
                 return Error(Conflict(message=f"User '{user_id}' already has role '{role_id}' in tenant '{tenant_id}'"))
 
             result_dict = {
--- a/backend/app/services/tenant.py
+++ b/backend/app/services/tenant.py
@@ -60,6 +60,7 @@ class TenantService:
             try:
                 tenant = await self._repository.create(tenant)
             except RepositoryConflictError:
+                await self._repository.rollback()
                 return Error(Conflict(message=f"Tenant '{name}' already exists"))
 
             result_dict = tenant.model_dump()


jamescrowley321:

None.

### WARN (should fix)

- **W1: `sync_role_assignment` passes canonical UUID as Descope `loginId`, causing silent sync failure**
  - Location: `backend/app/services/role.py` `assign_role_to_user()` calls `sync_role_assignment(user_id=user_id, tenant_id=tenant_id, role_id=role_id, data={"role_name": role.name})`. The `data` dict includes `role_name` but omits `login_id`. In `backend/app/services/adapters/descope.py:174-175`, the adapter falls back to `str(user_id)` (a canonical UUID) for the Descope `loginId` parameter.
  - Descope's `/v1/mgmt/user/update/role/add` endpoint requires `loginId` to be the user's email or phone, not an opaque UUID. The existing `DescopeManagementClient.resolve_login_id()` at `backend/app/services/descope.py:119-132` exists precisely for this translation, but is never called from the new code path. The existing router at `backend/app/routers/roles.py:70-71` handles this correctly by calling `resolve_login_id` before `assign_roles`.
  - **Impact:** Every `assign_role_to_user` call commits the assignment to the canonical DB, but the Descope sync will fail (user not found by UUID as loginId). Since sync failures are swallowed and `Ok(...)` is returned, this creates persistent divergence: the canonical store records the role assignment while Descope does not. JWTs issued by Descope will not carry the newly assigned role, so `require_role()` middleware enforcement will deny the user access they should have. The failure mode is denial of legitimate access, not privilege escalation -- no attacker can exploit this for unauthorized access.
  - **Fix:** Pass the user's email or Descope loginId in the `data` dict, e.g. `data={"role_name": role.name, "login_id": user_email}`, after resolving the user's email from the canonical User record or via `resolve_login_id()`.

### INFO (acceptable risk)

- **I1: Sync failures intentionally swallowed (write-through with tolerated divergence).** All three new services (`role.py`, `permission.py`, `tenant.py`) catch `SyncError` from the adapter, log at WARNING level, and return `Ok(...)`. This is an explicit architectural decision matching the existing `UserService` pattern. Acceptable provided the sync data is correct (see W1) and a reconciliation mechanism exists or is planned.

- **I2: Tenant isolation preserved in all queries.** `TenantRepository.get_users_with_roles()` filters by explicit `tenant_id` in the WHERE clause. `RoleRepository.get_by_name()` and `list_by_tenant()` scope by `tenant_id`. `UserTenantRoleRepository.list_by_user_tenant()` requires both `user_id` and `tenant_id`. No cross-tenant data leak vector exists.

- **I3: No new HTTP endpoints exposed.** This diff adds service and repository layers only. The dependency factories (`get_role_service`, `get_permission_service`, `get_tenant_service`) are defined but not yet wired to any router. No new attack surface is created. The existing middleware stack (TokenValidationMiddleware, RBAC via `require_role`) will need to protect these services when routers are added.

- **I4: No SQL injection risk.** All four new repositories use SQLAlchemy ORM query builder (`sa.select`, `sa.delete`, `.where()` with bound parameters). No raw SQL strings anywhere. Consistent with `UserRepository` which also properly escapes LIKE patterns via `_escape_like()`.

- **I5: No credential exposure in logs or error responses.** Sync failure logs include only entity IDs (UUIDs) and operation names. `RepositoryConflictError` raw strings (which may contain DB-level details) are caught at the service layer and replaced with typed `Conflict` errors containing clean user-facing messages.

- **I6: OTel span attributes contain entity names but no PII.** Span attributes include UUIDs and entity names (`role.name`, `permission.name`, `tenant.name`). These are not PII and are consistent with existing patterns.

- **I7: `list_all()` methods lack pagination.** `PermissionRepository.list_all()` and `TenantRepository.list_all()` return all records without limit/offset. Not exploitable today since no router consumes them, but should be paginated before exposure to HTTP traffic.

### Summary

- BLOCK: 0 | WARN: 1 | INFO: 7
- Overall: PASS

The single WARN identifies a correctness bug in the Descope sync path where `loginId` falls back to a canonical UUID instead of the user's email. The failure mode is denial of access (user does not gain the role they should have), not privilege escalation. Fix before the sync path is relied upon in production to prevent broken authorization grants and silent canonical/IdP divergence.

PR #206, opened 2026-04-06
feat: Unit + Integration Tests for Identity Domain (Story 2.4)

jamescrowley321:
## Summary
- Add repository unit tests (58 tests) for all 5 repository classes against real Postgres via testcontainers
- Add missing sync failure handling tests for all domain services (UserService, RoleService, PermissionService, TenantService)
- Add SyncFailed → 207 Multi-Status router response tests for all 4 identity routers
- Add integration lifecycle tests (create → get → update → delete) for all domain services with NoOpSyncAdapter against real Postgres
- Fix integration conftest.py to use subprocess for Alembic migrations and NullPool for cross-event-loop safety

## Story
Refs #147
Part of PRD 5: Canonical Identity Domain Model

## Chained PR
Based on #205 — merge first

## Review Findings Addressed
- Security: 0 BLOCK, 0 WARN — PASS
- Blind Hunter: 1 MUST FIX (iteration 1), 0 MUST FIX (iteration 2) — all resolved
- Edge Cases: 3 CRASH + 4 DEGRADED paths identified — all pre-existing production code, not in diff
- Acceptance: 5 PASS, 0 FAIL, 0 PARTIAL — all ACs satisfied

## Test plan
- [x] Unit tests pass (1127 tests)
- [x] Integration tests pass (22 tests)
- [x] Lint passes (`make lint`)
- [x] Independent review agents passed (2 iterations)
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #148, opened 2026-03-31
Epic 2 Story 2.5: E2E Tests + Regression


jamescrowley321:
## Story 2.5: E2E Tests + Regression

**Epic:** 2 — Identity & Access Administration

As a QA engineer,
I want Playwright E2E tests verifying the identity API after the service layer change,
So that we confirm the Postgres backing is invisible to API consumers.

### Acceptance Criteria

**AC-2.5.1: Regression**
**Given** the Playwright E2E test suite
**When** \`make test-e2e\` runs
**Then** ALL existing E2E tests pass without modification

**AC-2.5.2: Identity CRUD E2E**
**Given** new E2E tests for identity CRUD
**When** they run against the full stack
**Then** user CRUD operations succeed via the API
**And** role CRUD operations succeed
**And** permission CRUD operations succeed
**And** tenant operations succeed
**And** 3-tier auth enforced: unauthenticated (401), wrong role (403), admin (success)

**AC-2.5.3: Frontend compatibility**
**Given** the frontend application
**When** it interacts with the backend API
**Then** all UI functionality works identically (API contracts unchanged)

### Technical Notes
- Follow existing E2E patterns from PR #94 (3-tier auth)
- Use \`auth_api_context\` (OIDC client credentials) and \`admin_api_context\` (session token)
- Must verify the API shape is identical — Postgres backing should be invisible
- The onion architecture refactor is internal — E2E tests should not need to know about repository/service split

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #207, opened 2026-04-06
feat: E2E Tests + Regression for Identity CRUD (Story 2.5)

jamescrowley321:
## Summary
- Add comprehensive E2E tests for identity CRUD operations (users, roles, permissions, tenants)
- Verify 3-tier auth enforcement: unauthenticated (401), non-admin (403), admin (success)
- Validate response shapes include canonical fields (id, created_at, updated_at)
- Regression verification: existing E2E tests unaffected by onion architecture refactor

## Story
Refs #148

## Chained PR
Based on #206 (merged into story-2.3 branch) — merge story-2.3 first

## Review Findings Addressed
- Security: 0 BLOCK, 2 WARN — out of scope (existing helpers/auth.py)
- Blind Hunter: 0 MUST FIX, 1 SHOULD FIX, 2 NITPICK — all resolved
- Edge Cases: 0 findings
- Acceptance: 4/4 PASS, 0 FAIL, 0 PARTIAL

## Verification Evidence
- Lint: pass (ruff check + format)
- Unit tests: 1127 passed
- Frontend tests: 86 passed (9 suites)

## Test plan
- [x] Unit tests pass (`make test-unit` — 1127 passed)
- [x] Frontend tests pass (`make test-frontend` — 86 passed)
- [ ] CI passes
- [ ] E2E tests pass (`make test-e2e` — requires live Descope environment)

### DIFF
--- a/backend/tests/e2e/test_identity_crud_e2e.py
+++ b/backend/tests/e2e/test_identity_crud_e2e.py
@@ -473,9 +473,9 @@ class TestMemberCrud:
                 pytest.skip("Invite returned 207 (sync failed) — canonical fields not in Problem Detail body")
             assert "user" in body, f"Missing 'user' key in invite response: {body.keys()}"
             user = body["user"]
-            created_user_id = str(user["id"])
             assert "id" in user, f"Missing 'id' in user response: {user}"
             assert "created_at" in user, f"Missing 'created_at' in user response: {user}"
+            created_user_id = str(user["id"])
         finally:
             if created_user_id:
                 with contextlib.suppress(Exception):
@@ -579,3 +579,7 @@ class TestTenantOperations:
         body = resp.json()
         assert "tenant_id" in body
         assert "tenant" in body, f"Missing 'tenant' key in current tenant response: {body.keys()}"
+        tenant = body["tenant"]
+        assert "id" in tenant, f"Missing 'id' in tenant response: {tenant}"
+        assert "created_at" in tenant, f"Missing 'created_at' in tenant response: {tenant}"
+        assert "updated_at" in tenant, f"Missing 'updated_at' in tenant response: {tenant}"


jamescrowley321:

No blocking findings.

---

### WARN (should fix)

- [UNLIKELY] `backend/tests/e2e/helpers/auth.py:239-248` — Admin session JWT payload decoded and printed in full during test runs

  `get_admin_session_token()` prints all decoded JWT claims including `sub`, `dct`, `tenants`, `iss`, and `aud` to stdout unconditionally. In CI environments, stdout is captured in build logs that are often retained and accessible beyond the lifetime of the test run. The token itself is already expired by the time logs are reviewed (short-lived session JWT), but the tenant ID, user subject ID, and role structure in the claims are persistent metadata.

  Mitigation: Gate the debug print behind a `DEBUG` or `E2E_VERBOSE` env flag, or redact tenant and subject values before logging.

- [UNLIKELY] `backend/tests/e2e/helpers/auth.py:193-234` — Temporary privileged access key not guaranteed to be deleted on key creation failure

  `get_admin_session_token()` creates a temporary access key with `owner` + `admin` roles in the target tenant. Cleanup runs in a `finally` block wrapping only the exchange call (lines 215-234), which is correct for the exchange step. However, if `resp.raise_for_status()` on key creation (line 205) raises before `key_id` is populated, the `finally` block in the outer scope is never entered — there is no outer `try/finally` ensuring deletion when creation partially succeeds (key created, but ID not captured from the response). In that scenario the privileged key persists until it expires or is manually removed.

  Mitigation: Wrap the entire key-create + exchange + delete sequence in a single `try/finally` block, and only attempt deletion when `key_id` is non-empty.

---

### INFO (acceptable risk)

- `backend/tests/e2e/test_identity_crud_e2e.py:54-56` — Non-2xx responses print full response body via `resp.text()`

  In E2E tests against a live backend, error responses may include internal detail strings. Acceptable in a test context — the output only appears locally or in CI logs that already require the same credentials to access.

- `backend/tests/e2e/test_identity_crud_e2e.py:67` — Nil UUID (`000...000`) used for unauthenticated/non-admin path tests

  Auth enforcement runs before any resource lookup, so the server never queries this ID. No lookup-bypass risk.

- `backend/tests/e2e/test_identity_crud_e2e.py:471-472` — Test accepts HTTP 207 as a success code from `/members/invite`

  The 207 path is documented as "Postgres OK, Descope sync failed" and is a deliberate production design. The test correctly acknowledges the partial-success case; the endpoint still enforces `require_role("owner", "admin")` regardless of the sync outcome.

- `backend/tests/e2e/conftest.py:33` — `ignore_https_errors: True` in Playwright browser context args

  Suppresses TLS cert errors for local/CI testing against localhost. Does not affect API-level auth tests which use `APIRequestContext` directly and do not carry this flag.

- `backend/tests/e2e/helpers/auth.py:147-169` — Access token reused as `id_token` in injected OIDC session state

  A test-harness convenience for injecting an authenticated browser session without a real authorization code flow. Entirely scoped to test setup; does not reflect or affect any production auth path.

---

### Summary

- BLOCK: 0 | WARN: 2 | INFO: 5
- Overall: PASS

This diff adds only a test file. No production code is changed. The 3-tier auth model (401/403/200) is actively verified end-to-end. The two WARN items are operational hygiene issues in CI log exposure and a narrow key-cleanup gap in the test helper; neither is exploitable against production users.

Issue #161, opened 2026-03-31
Epic 1 Story 1.1: Create Tyk Configuration Directory


jamescrowley321:
## Story 1.1: Create Tyk Configuration Directory

As a developer setting up the gateway,
I want a `tyk/` directory with the gateway configuration file and placeholder directories,
So that all Tyk configuration is version-controlled and self-contained.

**Acceptance Criteria:**

- [ ] `tyk/` directory exists at the repo root containing `tyk.conf`, `apps/` (empty), `policies/` (empty), and `middleware/` (empty placeholder)
- [ ] `tyk.conf` configures `use_db_app_configs: false` (file-based mode), Redis connection to `tyk-redis` service on port 6379, and file-based policy source pointing to `/opt/tyk-gateway/policies/policies.json` (FR-1, FR-2)
- [ ] `tyk.conf` does NOT hardcode the gateway secret — it uses the `TYK_GW_SECRET` environment variable passed through Docker Compose (FR-7, NFR-4)
- [ ] `tyk/policies/policies.json` contains a default policy skeleton (empty policy array or minimal default)
- [ ] A `.gitkeep` file in `tyk/middleware/` preserves the empty directory for future plugin use
- [ ] Removing the `tyk/` directory has no impact on backend or frontend code (NFR-12)

**FRs covered:** FR-1, FR-2, FR-7

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 2 (API Gateway) shipped in full on 2026-04-11/12 via PRs #210-#239.

PR #212, opened 2026-04-07
feat: add Tyk gateway configuration directory

jamescrowley321:
## Summary
- Create `tyk/` directory with file-based gateway configuration (`tyk.conf`), empty API definition placeholder (`apps/.gitkeep`), default policy skeleton (`policies/policies.json`), and middleware placeholder (`middleware/.gitkeep`)
- Gateway secret uses `${TYK_GATEWAY_SECRET}` placeholder per arch doc — overridden at runtime via `TYK_GW_SECRET` env var in Docker Compose (Story 1.3)
- Redis configured for `tyk-redis:6379`, file-based policy source

## Story
Refs #161
Part of Epic 1: Tyk Gateway Integration

## Review Findings Addressed
- Security: 0 BLOCK, 2 WARN — all deferred (require running Tyk, Story 1.3)
- Blind Hunter: 2 MUST FIX, 3 SHOULD FIX, 2 NITPICK — overridden by Sentinel (context-aware) to PASS
- Edge Cases: 0 findings (config-only, no executable code paths)
- Acceptance: 5 PASS, 0 FAIL, 1 PARTIAL (AC-4 wording ambiguity, not a code defect)
- Red Team: 2 HIGH, 1 MEDIUM, 1 LOW — all require running Tyk (deferred to Story 1.3)

## Test plan
- [x] All 909 existing unit tests pass
- [x] JSON validation on tyk.conf and policies.json
- [x] No existing files modified (purely additive)
- [ ] CI passes

### DIFF
--- a/tyk/tyk.conf
+++ b/tyk/tyk.conf
@@ -1,6 +1,6 @@
 {
   "listen_port": 8080,
-  "secret": "",
+  "secret": "${TYK_GATEWAY_SECRET}",
   "template_path": "/opt/tyk-gateway/templates",
   "tyk_js_path": "/opt/tyk-gateway/js/tyk.js",
   "middleware_path": "/opt/tyk-gateway/middleware",


jamescrowley321:

_None._

### WARN (should fix)
- [UNLIKELY] `tyk/tyk.conf:8` — Redis has no password configured in the Tyk storage block.
  The existing `redis` service in `docker-compose.yml` is launched with `--requirepass ${REDIS_PASSWORD:-changeme}`, but `tyk.conf` carries no `password` field in its `storage` stanza. When the Tyk service is wired up in Story 1.3, it will fail to authenticate against Redis unless the password is added here (or via the `TYK_GW_STORAGE_PASSWORD` env-var override). This is not exploitable today because no Tyk container exists, but it is a latent misconfiguration that will either cause a startup failure or — if the Redis instance is somehow misconfigured to allow unauthenticated connections — could let Tyk read/write session keys without credential validation.
  Mitigation: Add `"password": "${TYK_GW_STORAGE_PASSWORD}"` to the `storage` block and supply the env var through Docker Compose in Story 1.3.

- [UNLIKELY] `tyk/tyk.conf` — `hash_key_function` is `murmur64`, a non-cryptographic hash.
  Tyk uses this to hash API keys stored in Redis. MurmurHash is fast and collision-prone under adversarial input. An attacker who can observe or enumerate Redis key hashes and craft API keys that produce the same hash could cause a key collision, potentially accessing another user's session. This is theoretical at present because the attack requires Redis read access, which is not externally exposed. Acceptable for local dev; for shared/production deployments prefer `bcrypt` or `sha256` (Tyk supports both).
  Mitigation: Document the choice; switch to `"hash_key_function": "sha256"` before any non-local deployment.

### INFO (acceptable risk)
- `tyk/tyk.conf:2` — `"secret": "${TYK_GATEWAY_SECRET}"` is a literal placeholder string, not a real secret. Tyk does not perform shell-variable expansion in its JSON config; the actual value is injected at runtime via the `TYK_GW_SECRET` environment variable (Docker Compose env block, Story 1.3). The placeholder is self-documenting and will not be treated as a valid gateway secret by any running Tyk instance. No credential exposure.

- `tyk/tyk.conf` — `"enable_analytics": false` disables Tyk's request analytics. This means no gateway-level audit trail for API calls in the Tyk layer. Acceptable for the current local-dev story; should be revisited before a shared or production deployment where forensic logging matters.

- `tyk/policies/policies.json` — Empty policy set (`"Data": {}`). No policies means no Tyk-managed rate-limit or access-control policies are active. This is intentional for the scaffold story; policies are populated in a later story. Not exploitable because no API definitions or keys exist yet either.

- `tyk/apps/.gitkeep`, `tyk/middleware/.gitkeep` — Empty placeholder directories committed to version control. No security impact. Standard scaffolding practice.

- Gitleaks (`/.gitleaks.toml`) does not allowlist the new `tyk/` path. The placeholder string `${TYK_GATEWAY_SECRET}` is unlikely to trigger a gitleaks rule (it is not a high-entropy string), but confirm by running `gitleaks detect` locally before merge to avoid a surprise CI failure.

### Summary
- BLOCK: 0 | WARN: 2 | INFO: 5
- Overall: PASS

Issue #151, opened 2026-03-31
Epic 3 Story 3.3: Redis Pub/Sub + Cache Invalidation


jamescrowley321:
## Story 3.3: Redis Pub/Sub + Cache Invalidation

**Epic:** 3 — Inbound Sync & Reconciliation
**FRs:** FR29
**Architecture:** D9

As a developer,
I want Redis pub/sub events on canonical data changes,
So that caches are invalidated immediately when identity data changes.

### Acceptance Criteria

**AC-3.3.1: Publish on write**
**Given** a canonical write operation (create, update, delete on user/role/permission/tenant)
**When** the Postgres write commits
**Then** a cache invalidation event is published to Redis pub/sub channel `identity:changes`
**And** the event includes: entity type, entity ID, operation, tenant_id, timestamp

**AC-3.3.2: Redis unavailability**
**Given** Redis is unavailable
**When** a canonical write occurs
**Then** cache invalidation events are silently dropped (fire-and-forget)
**And** canonical writes are NOT affected

**AC-3.3.3: Subscriber documentation**
**Given** a subscriber listening on `identity:changes`
**When** an event is received
**Then** the relevant cache entries can be invalidated (key pattern documented)

### Technical Notes
- Redis is already in Docker Compose from Story 1.1
- Fire-and-forget: reconciliation is the durability guarantee, not pub/sub
- Event schema should be documented for Epic 4's Redis cache consumer

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #214, opened 2026-04-07
feat: Redis Pub/Sub Cache Invalidation (Story 3.3)

jamescrowley321:
## Summary
- Add `CacheInvalidationPublisher` with fire-and-forget Redis pub/sub on `identity:changes` channel
- Publish entity change events (entity_type, entity_id, operation, tenant_id, timestamp) after every canonical write across all 7 domain services (User, Role, Permission, Tenant, InboundSync, Reconciliation)
- Redis unavailability silently degrades — canonical writes are never affected (AC-3.3.2)
- Singleton publisher initialized in FastAPI lifespan with graceful degradation and URL scheme validation
- 29 new unit tests covering publisher, singleton lifecycle, and per-service integration

## Story
Refs #151

## Chained PR
Based on #211 — merge first

## Review Findings Addressed
- Security: 0 BLOCK, 2 WARN (W6 URL scheme validation, W7 singleton sentinel) — all resolved
- Blind Hunter: 4 MUST FIX, 7 SHOULD FIX, 4 NITPICK — all resolved
- Edge Cases: 3 paths — all resolved
- Acceptance: 2 PASS, 1 PARTIAL (publisher tests added for all services) — all resolved
- Red Team: no findings

## Test plan
- [x] Unit tests pass (`make test-unit`) — 1323 tests
- [x] Independent review agents passed (2 iterations)
- [ ] CI passes


jamescrowley321:
## Review: Blind Hunter

### MUST FIX

- [`backend/app/services/inbound_sync.py:329`] Wrong operation label on create/link publish — uses `operation="sync"` for the create-or-link branch, but the correct operation for a newly-created user should be `"create"` (which IS used at line 319 for the existing-user update path). The variable `created` is available and distinguishes the two cases; consumers relying on `"create"` will miss cache invalidation for newly created users flowing through this path.

- [`backend/app/services/inbound_sync.py:349`] Wrong operation label on user deletion — `user.deleted` webhook deactivates a user but publishes `operation="sync"` instead of `operation="deactivate"`. The schema documents `"deactivate"` as a valid operation. Any consumer keyed on `"deactivate"` will never receive this event and will not invalidate caches for deactivated users.

- [`backend/uv.lock` (~line 1636, redis package entry)`] `redis` lock entry lists `pyjwt` as a dependency — `redis` has no dependency on `pyjwt`. This is a corrupt lockfile entry, likely a hand-edit or merge artifact. In production it is only harmless if `pyjwt` remains in the dependency tree via another package. If `pyjwt` is removed, `uv` would fail to install the environment. The lockfile must be regenerated with `uv lock` to produce a valid, reproducible entry.

- [`backend/app/main.py:95-103`] Leaked Redis connection when ping fails after TCP connect — on ping failure the `except Exception` block sets `redis_client = None` and discards the reference, but the `aioredis` client object that was allocated at line 98 is never closed. If the connection was established before the ping exception (e.g., AUTH failure, NOAUTH, wrong database), the underlying socket is abandoned without calling `aclose()`. Add `await redis_client.aclose()` in the except block before nulling the reference.

### SHOULD FIX

- [`backend/app/services/cache_invalidation.py:267-275`] `get_cache_publisher()` allocates a fresh `CacheInvalidationPublisher()` instance every time it is called before `init_cache_publisher` runs. Any caller that caches the returned instance before lifespan completes will hold a no-op publisher forever, even after the real one is initialised. A single module-level no-op sentinel (`_NOOP_PUBLISHER = CacheInvalidationPublisher()`) returned from the fallback branch would be safer and eliminates repeated allocation.

- [`backend/app/services/role.py:489`] `update_role` publishes without `tenant_id` — the call has no `tenant_id` kwarg so it defaults to `None`. `create_role` (line 477) and `assign_role_to_user` (line 499) correctly pass `tenant_id`. Roles can be tenant-scoped; consumers filtering on `tenant_id` will miss cache invalidation for tenant-scoped role updates.

- [`backend/app/services/user.py:653-656`] Remove-user-from-tenant publishes `operation="update"` — removing a tenant membership is a removal/unassignment, and the documented schema lists `"unassign"` as a valid operation. `RoleService.unassign_role_from_user` correctly uses `"unassign"`. Using `"update"` here creates inconsistency; consumers cannot distinguish membership removal from a field edit.

- [`backend/app/main.py:115-120`] `shutdown_cache_publisher()` and `redis_client.aclose()` share a single `try/except` block. Although `shutdown_cache_publisher()` currently cannot raise (it is a `global _publisher = None` assignment), if it ever does, `redis_client.aclose()` will be skipped because the except clause absorbs the exception and continues. The two operations should be in separate try/finally blocks to guarantee both execute.

- [`backend/app/services/cache_invalidation.py:230-243`] `publish_batch` constructs the event dict inline and duplicates the `datetime.now(timezone.utc).isoformat()` call that exists in `_build_event`. The batch event diverges from the standard schema independently. If the timestamp format or timezone handling is ever changed in `_build_event`, the batch path will silently produce inconsistent timestamps. Extract a shared `_now_iso()` helper or use `_build_event` for the common fields.

- [`backend/tests/unit/test_cache_invalidation.py:802-819`] `TestSingletonLifecycle` tests mutate module-level global state without using fixtures to guarantee cleanup. If `test_init_and_get_returns_same_instance` fails before reaching its `shutdown_cache_publisher()` cleanup call, all subsequent tests in the file inherit a dirty singleton. Use `setup_method`/`teardown_method` or a pytest fixture with `autouse=True` to guarantee the singleton is reset regardless of test outcome.

- [`backend/app/services/cache_invalidation.py:280-282`] `init_cache_publisher` performs a bare `global _publisher` assignment with no guard against double-initialisation. If called twice (e.g., in tests that forget to call `shutdown_cache_publisher` between runs), the first publisher's Redis client reference is silently replaced without being closed. The old client leaks. Add a warning log or raise if `_publisher` is already set to a non-None value.

### NITPICK

- [`backend/app/main.py:74`] `import os` added solely for `os.getenv("REDIS_URL")`. The project presumably has a settings/config object that validates env vars at startup. Using raw `os.getenv` bypasses any such validation layer and is inconsistent with how the rest of the app likely reads configuration.

- [`backend/app/services/cache_invalidation.py:203`] Constructor parameter is named `redis_client` but the stored attribute is `self._redis` — the shortened name is slightly inconsistent and marginally harder to search for.

- [`backend/app/services/cache_invalidation.py:192`] `CHANNEL = "identity:changes"` is exported implicitly. If downstream consumers import this constant, they are coupling to the internal module path. Worth noting in the module docstring that `CHANNEL` is part of the stable contract.

- [`backend/app/services/inbound_sync.py:305`] `publisher: CacheInvalidationPublisher | None = None` default is consistent with all other services in this PR, but the `if self._publisher:` guard scattered through business logic is noise. A null-object pattern (always inject a real or no-op publisher, never `None`) would eliminate all seven guard clauses across the six service files.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `backend/app/services/inbound_sync.py:306` | `_handle_user_deleted` webhook deactivates the user but publishes `operation="sync"` instead of `"deactivate"`; the documented schema lists `"deactivate"` as a distinct operation | Change to `operation="deactivate"` | [WRONG] Downstream cache consumers dispatching on operation type see a sync event, not a deactivation; deactivated users may not be invalidated correctly |
| `backend/app/main.py:87-89` | `shutdown_cache_publisher()` and `redis_client.aclose()` share one `try` block; if `shutdown_cache_publisher()` raises before `aclose()` is reached, the Redis connection is never closed | Split into two sequential `try/finally` blocks so `aclose()` always runs | [DEGRADED] Redis connection leaks on shutdown if `shutdown_cache_publisher` raises (currently cannot raise, but is a structural fragility) |
| `backend/app/services/cache_invalidation.py:114-122` | `get_cache_publisher()` called before `init_cache_publisher()` (e.g., after `shutdown_cache_publisher()` sets `_publisher=None` mid-flight) returns a fresh throwaway no-op instance per call rather than the singleton | Assign the fallback to `_publisher`, or document that only the lifespan path is supported | [DEGRADED] Each pre-init call returns a distinct throwaway object; services injected at DI time hold a different no-op instance than any future call; behaviorally safe but identity is inconsistent and can mask misconfiguration in tests |

### Summary
- Unhandled paths found: 3

jamescrowley321:
## Review: Acceptance Auditor

### PASS

- [AC-3.3.2] Redis unavailability — all exceptions in `publish()` and `publish_batch()` are caught with a bare `except Exception`, logged as warnings, and never re-raised. The `repo.commit()` call precedes every `publish()` call in every service, so the Postgres write is always committed before Redis is attempted. A `None` redis client (set when `REDIS_URL` is absent or Redis connection fails at startup) causes `publish()` to return immediately without any network attempt. App startup at `backend/app/main.py:95-106` gracefully handles connection failure by setting `redis_client=None` before calling `init_cache_publisher`. Canonical writes are structurally unaffected.
  - Implemented at `backend/app/services/cache_invalidation.py:62-68` (None guard + except), `backend/app/main.py:92-106` (startup degradation)
  - Tested at `backend/tests/unit/test_cache_invalidation.py:56` (no-op when client is None), `backend/tests/unit/test_cache_invalidation.py:62` (exception swallowed, warning logged, no raise), `backend/tests/unit/test_cache_invalidation.py:111` (batch swallows exception)

- [AC-3.3.3] Subscriber documentation — the module-level docstring at `backend/app/services/cache_invalidation.py:1-24` documents: the channel name (`identity:changes`), the full JSON event schema with allowed values for each field, and key patterns subscribers should use to identify affected records and scope invalidation to a tenant. The schema is validated by tests that deserialize actual published payloads and assert all fields are present with correct values.
  - Implemented at `backend/app/services/cache_invalidation.py:9-23`
  - Tested at `backend/tests/unit/test_cache_invalidation.py:32-54` (channel name, all schema fields including timestamp and null tenant_id verified against a live publish call)

### FAIL

- None

### PARTIAL

- [AC-3.3.1] Publish on write — `CacheInvalidationPublisher.publish()` is called after `repo.commit()` in all four canonical entity services (user, role, permission, tenant) and for all write operations. The event includes all required fields (entity_type, entity_id, operation, tenant_id, timestamp). Channel is `identity:changes`.
  - Implemented at: `backend/app/services/user.py:83,153,200,243,284`; `backend/app/services/role.py:101,168,232,299,332,368`; `backend/app/services/permission.py:74,140,174`; `backend/app/services/tenant.py:73`
  - **What's done:** Implementation is present and correct in all four services. `test_user_service.py:TestCacheInvalidationPublishing` (lines 478-532) covers UserService create + deactivate publish assertions, and a guard test confirms no publish on failure. `test_cache_invalidation.py:TestPublish` validates the full event schema for the `publish()` method itself.
  - **What's missing:** `backend/tests/unit/test_role_service.py`, `backend/tests/unit/test_permission_service.py`, and `backend/tests/unit/test_tenant_service.py` contain zero references to `CacheInvalidationPublisher` or `publisher`. There are no tests that would fail if the `if self._publisher: await self._publisher.publish(...)` blocks were removed from `role.py`, `permission.py`, or `tenant.py`. Three of four canonical entity types lack service-level publish tests.

### SCOPE CREEP

- `backend/app/services/inbound_sync.py:318-319`, `:329`, `:339`, `:349` — `InboundSyncService` publishes cache events for flow-sync and webhook-triggered user mutations (operation `"sync"`). AC-3.3.1 specifies "canonical write operations (create, update, delete on user/role/permission/tenant)." Inbound sync is a secondary write path not listed in the AC's entity/operation enumeration. The publisher is wired via `backend/app/dependencies/identity.py:56`. Not traceable to any AC.

- `backend/app/services/reconciliation.py:156-157` — `ReconciliationService.run()` calls `publish_batch()` after a reconciliation pass with changes. Reconciliation is not a canonical write operation per the spec. Additionally, `publish_batch()` emits events with `entity_type="batch"` and `entity_id="reconciliation"`, which fall outside the documented schema in AC-3.3.3 (entity_type is specified as `user | role | permission | tenant`). Wired via `backend/app/dependencies/identity.py:57`. Not traceable to any AC.

### Architecture Violations

- None identified. `CacheInvalidationPublisher` is injected into domain services via constructor, consistent with D4 (constructor injection, onion architecture). The publisher is always called after `await self._repository.commit()`, preserving the D7 write-through Postgres-first semantic. No publisher calls occur in the repository layer (inner) or adapter layer (outer). The singleton pattern mirrors the existing `descope.py` module pattern. The Redis import is gated behind `TYPE_CHECKING` to avoid a hard load-time import.

### Summary

- Total ACs: 3
- Pass: 2 | Fail: 0 | Partial: 1

The partial is AC-3.3.1: implementation is complete and correct across all four canonical entity services, but three of the four service test files (role, permission, tenant) have no publisher assertions. Removing the publish calls from those three services would not be caught by the test suite.

jamescrowley321:

- none

### WARN (should fix)

**From prior review (Story 3.1 — still open):**

**New findings — Story 3.3 (cache invalidation):**

- **W6** [`backend/app/main.py:93-98`] Redis URL is consumed without scheme validation. `aioredis.from_url(redis_url)` accepts any URL scheme the `redis` library supports, including `unix://`. If `REDIS_URL` is misconfigured or injected (e.g., via a compromised secrets manager), the application will silently attempt to connect to a non-Redis endpoint. The graceful-degradation catch block swallows the error, which is intentional, but a badly formed URL pointing at an internal metadata endpoint (e.g., `http://169.254.169.254/...`) would cause a connection attempt to that host at startup without any scheme allowlist.
  Mitigation: Validate `redis_url` scheme before calling `aioredis.from_url()` — e.g., assert scheme in `{"redis", "rediss"}`.

- **W7** [`backend/app/services/cache_invalidation.py:267-275`] `get_cache_publisher()` has a singleton gap: if called before `init_cache_publisher()` (e.g., in tests that import the module without going through app lifespan), it returns a throwaway no-op `CacheInvalidationPublisher()` that is not stored in `_publisher`. A consumer that caches this returned reference before init completes will hold a permanently no-op publisher and silently miss all cache invalidation events after Redis connects. The security implication is that downstream caches may serve stale identity or permission data indefinitely with no indication of the misconfiguration.
  Mitigation: Document init ordering requirement; add a warning log if `get_cache_publisher()` is called before `init_cache_publisher()`.

### INFO (acceptable risk)

- **I8** No SQL injection risk. All data flows through SQLAlchemy ORM parameterized queries.

- **I10** Cache invalidation events contain no credentials or PII. Published events carry only UUIDs and enum-like strings. The `stats` dict in `publish_batch` contains only server-side integer counters — no user-supplied data flows into the Redis channel payload. A compromised Redis subscriber learns only that a mutation occurred on a given entity type/ID, not the content of the change.

- **I11** Redis auth is configured in docker-compose. `REDIS_URL` includes password authentication; Redis port 6379 is bound to `127.0.0.1` only, limiting exposure to the Docker network. The default `changeme` password is documented as overridable via `REDIS_PASSWORD`.

- **I12** Cache invalidation failure is genuinely fire-and-forget. All `publish()` and `publish_batch()` calls are wrapped in `try/except Exception`. Failures never propagate to the service layer and never roll back committed DB transactions. A Redis outage cannot cause data inconsistency in the canonical store.

- **I13** No SSRF risk from Redis URL under normal operation. The URL is taken from a server-side environment variable, not from any user-supplied request parameter. (W6 above covers the env-injection edge case.)

- **I14** `publish_batch` stats integrity. The `stats` dict is initialized with hardcoded string keys and integer zero values; only `+= 1` increments modify it before publication. No user-controlled data reaches the published stats payload.

- **I15** `shutdown_cache_publisher()` correctly nulls the singleton before Redis client is closed (`main.py:87-89`). Subsequent calls to `get_cache_publisher()` after shutdown return a new no-op instance, preventing use-after-close of the Redis client.

### Summary
- Overall: PASS

jamescrowley321:
## Review: Red Team (Viper)

### Attack Surface

### Findings

- **Attack scenario**:

- **Attack scenario**:

- **Attack scenario**:

- **Attack scenario**:

### Summary

- Attack surface elements: 9
- Overall: PASS

backend/app/services/cache_invalidation.py
```diff
@@ -0,0 +1,133 @@
...
+        tenant_id: UUID | str | None = None,
+    ) -> None:
+        """Publish a single change event. Silent on failure (AC-3.3.2)."""
+        if self._redis is None:
+            return
+        event = self._build_event(entity_type, entity_id, operation, tenant_id)
+        try:
+            await self._redis.publish(CHANNEL, json.dumps(event))
+        except Exception:
+            logger.warning("Redis publish failed for %s/%s", entity_type, entity_id, exc_info=True)
```
jamescrowley321: Make this an error.

backend/app/services/cache_invalidation.py
```diff
@@ -0,0 +1,133 @@
...
+            "operation": operation,
+            "tenant_id": str(tenant_id) if tenant_id is not None else None,
+            "timestamp": datetime.now(timezone.utc).isoformat(),
+        }
+
+
+# ---------------------------------------------------------------------------
+# Module-level singleton (mirrors descope.py pattern)
+# ---------------------------------------------------------------------------
+_publisher: CacheInvalidationPublisher | None = None
```
jamescrowley321, line 105: This pattern is a code smell.

backend/app/services/cache_invalidation.py
jamescrowley321: Escalated to `logger.error` in e9598ed — AC-3.3.2 "never raised" contract preserved, `publish_batch` got the same treatment, docstrings and unit test assertions updated to match.

backend/app/services/cache_invalidation.py
jamescrowley321: Agreed — module-level singletons are code smell. Leaving as-is in this PR because T150 / #216 is a dedicated refactor task that rips out both this publisher singleton **and** the matching `descope.py` singleton and replaces them with `app.state` DI. Doing half the refactor here would either ship an inconsistent pattern or blow up the scope of story 3.3. The canonical loop will pick up #216 as its next task.

jamescrowley321:
Subsumed by #226 (merge commit 791d220), which landed the entire canonical chain (stories 2.2–4.4) on main in a single linearized commit. The story-3.3 Redis pub/sub work is included, plus the cache_invalidation log-level fix (warning → error) that was addressed in commit e9598ed was cherry-picked forward into #226. The `canonical/story-3.3-redis-pubsub` branch is being kept until cleanup is explicitly confirmed.

Issue #153, opened 2026-03-31
Epic 4 Story 4.1: IdP Link + Provider Config Service


jamescrowley321:
## Story 4.1: IdP Link + Provider Config Service

**Epic:** 4 — Multi-IdP Identity Linking
**FRs:** FR13-19
**Architecture:** NFR5

As a system operator,
I want to manage IdP links and provider configurations in the canonical store,
So that users can have linked identities across multiple providers and new IdPs can be registered.

### Acceptance Criteria

**AC-4.1.1: IdP link CRUD**
**Given** `IdPLinkRepository` and `IdPLinkService` following the onion architecture pattern
**When** `IdPLinkService.create_idp_link()` is called with user_id, provider_id, external_sub, external_email, optional metadata (JSONB)
**Then** an IdPLink record is created via repository with `linked_at` timestamp
**And** unique constraint prevents duplicate links (one user per provider)
**And** a single IdP identity cannot be linked to multiple canonical users

**Given** `get_user_idp_links()` called with user_id
**Then** all IdP links for that user are returned via repository

**Given** `delete_idp_link()` called
**Then** the link is removed via repository, `Ok(None)` returned

**AC-4.1.2: Provider config management**
**Given** `ProviderRepository` and `ProviderService` following the onion architecture pattern
**When** `register_provider()` is called with name, type, issuer_url, base_url, capabilities, config_ref
**Then** a Provider record is created via repository with `active=True`
**And** credentials NOT stored in Postgres — only `config_ref` path (NFR5)

**Given** `deactivate_provider()` called
**Then** provider's `active` flag set to False via repository

**Given** `get_provider_capabilities()` called
**Then** the capability list (`rbac`, `fga`, `scim`) is returned

**AC-4.1.3: Onion layer compliance**
**Given** all repositories and services created in this story
**Then** repositories contain NO business logic, NO OTel spans, NO adapter calls
**And** services contain NO direct SQLAlchemy imports
**And** DI factories compose: `session → repository → service(repository, adapter)`

### Technical Notes
- Provider types: descope, ory, entra, cognito, oidc (generic)
- config_ref format: Infisical path (e.g., `/identity/providers/descope/credentials`)
- Capabilities are a JSON array on the Provider model
- Follow same onion architecture pattern established in Epic 2 stories

jamescrowley321:
[...]

PR #219, opened 2026-04-07
feat: add IdP Link and Provider domain services

jamescrowley321:
## Summary

- Add `IdPLinkService` with create/get/delete operations for managing identity provider links, including user+provider validation, unique constraint enforcement, and OTel instrumentation
- Add `ProviderService` with register/deactivate/get-capabilities operations, config_ref-only storage (no credentials in Postgres), and idempotent deactivation
- Extend `IdPLinkRepository` and `ProviderRepository` with missing CRUD methods (get-by-id, delete, create, update) including proper rollback handling
- Add DI factories for both new services in `dependencies/identity.py`
- Comprehensive unit tests for both services and repository additions

Refs #153

## Review Findings Addressed

- **Blind Hunter**: 2 MUST FIX fixed (rollback in ProviderRepository, deactivate guard), 3 of 4 SHOULD FIX fixed (short-circuit, TOCTOU comment, test coverage), 1 accepted (commit exception matches codebase pattern)
- **Edge Case Hunter**: 4 findings all fixed (rollback alignment, rollback() method, deactivate guard)
- **Acceptance Auditor**: 12/13 PASS, 1 PARTIAL (non-blocking)
- **Sentinel**: 0 BLOCK, 2 WARN (tenant scoping + auth docs deferred to story 4.2 router layer)
- Delta re-review: all fixes verified correct, no regressions

## Test plan
- [x] Unit tests pass (all 1335 tests)
- [x] Lint passes
- [x] Independent review agents passed (blind, edge-case, acceptance, sentinel)
- [ ] CI passes

### DIFF
--- a/backend/app/dependencies/identity.py
+++ b/backend/app/dependencies/identity.py
@@ -5,6 +5,7 @@ AC-2.2.6: get_role_service(), get_permission_service(), get_tenant_service()
 with DescopeSyncAdapter wrapping the singleton DescopeManagementClient.
 AC-3.1.1: get_inbound_sync_service() wires repositories for inbound sync.
 AC-3.2.1: get_reconciliation_service() wires repositories + Descope client for reconciliation.
+AC-4.1.3: get_idp_link_service(), get_provider_service() for IdP link/provider domain services.
 """
 
 from __future__ import annotations
@@ -24,8 +25,10 @@ from app.repositories.user import UserRepository
 from app.services.adapters.descope import DescopeSyncAdapter
 from app.services.cache_invalidation import get_cache_publisher
 from app.services.descope import get_descope_client
+from app.services.idp_link import IdPLinkService
 from app.services.inbound_sync import InboundSyncService
 from app.services.permission import PermissionService
+from app.services.provider import ProviderService
 from app.services.reconciliation import ReconciliationService
 from app.services.role import RoleService
 from app.services.tenant import TenantService
@@ -158,3 +161,33 @@ async def get_reconciliation_service(
         provider_repository=provider_repository,
         publisher=get_cache_publisher(),
     )
+
+
+async def get_idp_link_service(
+    session: AsyncSession = Depends(get_async_session),
+) -> IdPLinkService:
+    """Build an IdPLinkService with its repositories.
+
+    AC-4.1.3: Wiring: AsyncSession -> IdPLinkRepository + UserRepository + ProviderRepository
+               -> IdPLinkService(repository, user_repository, provider_repository)
+    """
+    repository = IdPLinkRepository(session)
+    user_repository = UserRepository(session)
+    provider_repository = ProviderRepository(session)
+    return IdPLinkService(
+        repository=repository,
+        user_repository=user_repository,
+        provider_repository=provider_repository,
+    )
+
+
+async def get_provider_service(
+    session: AsyncSession = Depends(get_async_session),
+) -> ProviderService:
+    """Build a ProviderService with its repository.
+
+    AC-4.1.3: Wiring: AsyncSession -> ProviderRepository(session)
+               -> ProviderService(repository)
+    """
+    repository = ProviderRepository(session)
+    return ProviderService(repository=repository)
--- a/backend/app/repositories/idp_link.py
+++ b/backend/app/repositories/idp_link.py
@@ -25,6 +25,19 @@ class IdPLinkRepository:
     def __init__(self, session: AsyncSession) -> None:
         self._session = session
 
+    async def get(self, link_id: uuid.UUID) -> IdPLink | None:
+        """Fetch an IdP link by primary key. Returns None if not found."""
+        return await self._session.get(IdPLink, link_id)
+
+    async def delete(self, link_id: uuid.UUID) -> bool:
+        """Delete an IdP link by primary key. Returns True if deleted, False if not found."""
+        link = await self._session.get(IdPLink, link_id)
+        if link is None:
+            return False
+        await self._session.delete(link)
+        await self._session.flush()
+        return True
+
     async def get_by_provider_and_sub(self, provider_id: uuid.UUID, external_sub: str) -> IdPLink | None:
         """Fetch an IdP link by provider and external subject. Returns None if not found."""
         stmt = sa.select(IdPLink).where(
--- a/backend/app/repositories/provider.py
+++ b/backend/app/repositories/provider.py
@@ -29,12 +29,13 @@ class ProviderRepository:
         """Add a new provider to the session and flush to generate defaults.
 
         Raises RepositoryConflictError if a uniqueness constraint is violated.
-        Does NOT rollback — the caller (service layer) owns the transaction.
+        Rolls back the session before raising so it is not left in an invalid state.
         """
         self._session.add(provider)
         try:
             await self._session.flush()
         except IntegrityError as exc:
+            await self._session.rollback()
             raise RepositoryConflictError(str(exc)) from exc
         return provider
 
@@ -42,10 +43,12 @@ class ProviderRepository:
         """Flush pending mutations on a provider already attached to the session.
 
         Raises RepositoryConflictError if a uniqueness constraint is violated.
+        Rolls back the session before raising so it is not left in an invalid state.
         """
         try:
             await self._session.flush()
         except IntegrityError as exc:
+            await self._session.rollback()
             raise RepositoryConflictError(str(exc)) from exc
         return provider
 
@@ -72,3 +75,7 @@ class ProviderRepository:
     async def commit(self) -> None:
         """Commit the current transaction."""
         await self._session.commit()
+
+    async def rollback(self) -> None:
+        """Roll back the current transaction."""
+        await self._session.rollback()
--- a/backend/app/repositories/provider.py
+++ b/backend/app/repositories/provider.py
@@ -1,6 +1,6 @@
 """ProviderRepository — data access layer for canonical Provider model.
 
-Handles all SQLAlchemy queries for provider lookup.
+Handles all SQLAlchemy queries for provider CRUD.
 Contains NO business logic, NO OTel spans, NO adapter calls — data access only.
 """
 
@@ -9,9 +9,11 @@ from __future__ import annotations
 import uuid
 
 import sqlalchemy as sa
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from app.models.identity.provider import Provider, ProviderType
+from app.repositories.user import RepositoryConflictError
 
 
 class ProviderRepository:
@@ -22,6 +24,30 @@ class ProviderRepository:
 
     def __init__(self, session: AsyncSession) -> None:
         self._session = session
+
+    async def create(self, provider: Provider) -> Provider:
+        """Add a new provider to the session and flush to generate defaults.
+
+        Raises RepositoryConflictError if a uniqueness constraint is violated.
+        Does NOT rollback — the caller (service layer) owns the transaction.
+        """
+        self._session.add(provider)
+        try:
+            await self._session.flush()
+        except IntegrityError as exc:
+            raise RepositoryConflictError(str(exc)) from exc
+        return provider
+
+    async def update(self, provider: Provider) -> Provider:
+        """Flush pending mutations on a provider already attached to the session.
+
+        Raises RepositoryConflictError if a uniqueness constraint is violated.
+        """
+        try:
+            await self._session.flush()
+        except IntegrityError as exc:
+            raise RepositoryConflictError(str(exc)) from exc
+        return provider
 
     async def get_by_type(self, provider_type: ProviderType) -> Provider | None:
         """Fetch a provider by type. Returns the first match or None.
--- a/backend/app/services/idp_link.py
+++ b/backend/app/services/idp_link.py
@@ -1,0 +1,120 @@
+"""IdPLinkService — domain orchestration for canonical IdP link operations.
+
+Middle layer of onion architecture: orchestrates IdPLinkRepository, UserRepository,
+and ProviderRepository (data access). All methods return Result[T, IdentityError].
+OTel spans on every method.
+"""
+
+from __future__ import annotations
+
+import logging
+import uuid
+
+from expression import Error, Ok, Result
+from opentelemetry import trace
+
+from app.errors.identity import Conflict, IdentityError, NotFound
+from app.models.identity.user import IdPLink
+from app.repositories.idp_link import IdPLinkRepository
+from app.repositories.provider import ProviderRepository
+from app.repositories.user import RepositoryConflictError, UserRepository
+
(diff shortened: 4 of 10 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX

- [`backend/app/repositories/provider.py:119-120`] `ProviderRepository.create` and `ProviderRepository.update` catch `IntegrityError` and re-raise as `RepositoryConflictError` but do NOT rollback the session. After a SQLAlchemy `IntegrityError`, the session is in an invalid state — any subsequent operation on the same session will raise `sqlalchemy.exc.InvalidRequestError`. The callers (`register_provider`, `deactivate_provider`) return a clean `Error(Conflict(...))` result, but the underlying session is poisoned. If FastAPI middleware or any downstream code attempts to use that session (e.g., a second repository call in the same request), it will crash. The `UserRepository` in this same codebase calls `await self._session.rollback()` before re-raising — these new methods should do the same.

- [`backend/app/services/idp_link.py:83-84`] In `create_idp_link`, `link.model_dump()` is called, then `await self._repository.commit()` follows. If `commit()` raises an exception (database connection drop, serialization failure, deadlock), the exception propagates unhandled out of the `with tracer.start_as_current_span(...)` block with no error wrapping, no rollback, and no `Result.Error` return. The caller receives a raw exception instead of the `Result[dict, IdentityError]` they expect. Same issue exists in `ProviderService.register_provider` and `ProviderService.deactivate_provider`. Note: this matches the pattern in existing services, but `commit()` is the operation most likely to fail in production.

### SHOULD FIX

- [`backend/app/services/idp_link.py:88-100`] `get_user_idp_links` does not validate that the user exists before querying for their links. If `user_id` refers to a non-existent user, it silently returns `Ok([])` — an empty list. This is indistinguishable from a real user who has no IdP links. The `create_idp_link` method validates user existence, so this inconsistency could mask bugs in callers that pass bad user IDs.

- [`backend/app/services/provider.py:66-68`] In `register_provider`, the pre-check via `get_by_name` on line 66 and the `except RepositoryConflictError` fallback on line 80 form a dual-check pattern for the same uniqueness constraint. The TOCTOU race between `get_by_name` and `create` means both checks are needed, but neither code path documents why both exist. A future maintainer may remove one thinking it duplicates the other, re-introducing either the race condition (if the fallback is removed) or unnecessary object construction (if the pre-check is removed). Add a comment explaining the dual-check rationale.

- [`backend/app/services/provider.py:104-108`] `deactivate_provider` unconditionally sets `provider.active = False` and calls `update` + `commit` even when the provider is already inactive. The docstring says "Idempotent — already-inactive returns Ok" but it still performs a write (flush + commit) against the database on every call. This is a wasted round-trip. Should short-circuit with `if not provider.active: return Ok(provider.model_dump())` before the write path.

- [`backend/app/repositories/provider.py:123-132`] `ProviderRepository.update` only calls `flush()` — it does not verify the provider is attached to the session or that it has pending changes. If called with a detached or expunged provider instance, the flush is a silent no-op and the caller believes the update succeeded. In the current call site (`deactivate_provider`), the provider comes from `get()` so it is attached, but the method's public API makes no such guarantee.

### NITPICK

- [`backend/app/services/provider.py:49`] Parameter named `type` shadows the Python built-in `type()`. While it works, it prevents using `type()` inside this method body. Consider `provider_type` for clarity.

- [`backend/app/services/idp_link.py:75`] The field is named `metadata_` (with trailing underscore) in the model, and is passed as `metadata_=metadata` to the constructor. The public API of `create_idp_link` uses `metadata` (no underscore). This name mismatch between public API and internal model field is a maintenance trap — someone grepping for `metadata` usage won't find the model field, and vice versa.

- [`backend/app/repositories/provider.py:11`] `RepositoryConflictError` is imported from `app.repositories.user`. This error class is a general-purpose repository concern, not user-specific. Having every repository import from the user repository creates a false coupling. This pre-dates the diff but the diff extends the pattern to another file.

jamescrowley321:
## Review: Acceptance Auditor

### PASS

- [AC-4.1.1a] `IdPLinkService.create_idp_link()` accepts user_id, provider_id, external_sub, external_email, optional metadata (JSONB) and creates an IdPLink record via repository with `linked_at` timestamp (auto-set by model default) — implemented at `backend/app/services/idp_link.py:44-85`, tested at `backend/tests/unit/test_idp_link_service.py:51-71` (happy path), `backend/tests/unit/test_idp_link_service.py:122-141` (metadata)

- [AC-4.1.1b] Unique constraint prevents duplicate links (one user per provider) — enforced by DB constraint `uq_idp_links_user_provider` in model at `backend/app/models/identity/user.py:58`, caught as `RepositoryConflictError` in `backend/app/repositories/idp_link.py:56-60`, mapped to `Conflict` in `backend/app/services/idp_link.py:78-80`, tested at `backend/tests/unit/test_idp_link_service.py:105-120` (service) and `backend/tests/unit/repositories/test_idp_link_repository.py:73-80` (repository)

- [AC-4.1.1c] A single IdP identity cannot be linked to multiple canonical users — enforced by DB constraint `uq_idp_links_provider_external_sub` in model at `backend/app/models/identity/user.py:59`, tested at `backend/tests/unit/repositories/test_idp_link_repository.py:83-95`

- [AC-4.1.1d] `get_user_idp_links()` returns all IdP links for a user via repository — implemented at `backend/app/services/idp_link.py:87-100`, tested at `backend/tests/unit/test_idp_link_service.py:148-161` (list + empty), repository tested at `backend/tests/unit/repositories/test_idp_link_repository.py:134-153`

- [AC-4.1.1e] `delete_idp_link()` removes the link via repository — implemented at `backend/app/services/idp_link.py:102-120`, tested at `backend/tests/unit/test_idp_link_service.py:177-198` (success + not found), repository tested at `backend/tests/unit/repositories/test_idp_link_repository.py:116-131`

- [AC-4.1.2a] `register_provider()` accepts name, type, issuer_url, base_url, capabilities, config_ref and creates a Provider with `active=True` (model default) — implemented at `backend/app/services/provider.py:39-77`, tested at `backend/tests/unit/test_provider_service.py:51-71` (happy path), `backend/tests/unit/test_provider_service.py:73-85` (duplicate name), `backend/tests/unit/test_provider_service.py:87-100` (TOCTOU race), `backend/tests/unit/test_provider_service.py:102-117` (default capabilities). Repository create tested at `backend/tests/unit/repositories/test_provider_repository.py:15-24`

- [AC-4.1.2b] Credentials NOT stored in Postgres — only `config_ref` path (NFR5) — the Provider model at `backend/app/models/identity/provider.py:19-45` contains no credential fields; `config_ref` (Infisical path) is stored instead. Service docstring explicitly states this. Test at `backend/tests/unit/test_provider_service.py:68` verifies `config_ref` is returned in the result dict

- [AC-4.1.2c] `deactivate_provider()` sets `active` flag to False via repository — implemented at `backend/app/services/provider.py:79-101`, tested at `backend/tests/unit/test_provider_service.py:124-135` (success), `backend/tests/unit/test_provider_service.py:137-149` (idempotent for already-inactive), `backend/tests/unit/test_provider_service.py:151-159` (not found). Repository update tested at `backend/tests/unit/repositories/test_provider_repository.py:88-96`

- [AC-4.1.2d] `get_provider_capabilities()` returns the capability list (`rbac`, `fga`, `scim` etc.) — implemented at `backend/app/services/provider.py:103-119`, tested at `backend/tests/unit/test_provider_service.py:166-176` (with capabilities), `backend/tests/unit/test_provider_service.py:178-186` (empty list), `backend/tests/unit/test_provider_service.py:188-195` (not found)

- [AC-4.1.3a] Repositories contain NO business logic, NO OTel spans, NO adapter calls — verified by inspection: `backend/app/repositories/idp_link.py` and `backend/app/repositories/provider.py` contain only SQLAlchemy data access operations. No `tracer`, `opentelemetry`, or adapter imports found.

- [AC-4.1.3b] Services contain NO direct SQLAlchemy imports — verified by grep: `backend/app/services/idp_link.py` and `backend/app/services/provider.py` have zero `sqlalchemy` imports. All data access goes through repository method calls.

- [AC-4.1.3c] DI factories compose: `session -> repository -> service(repository, ...)` — implemented at `backend/app/dependencies/identity.py:166-193` (`get_idp_link_service` wires 3 repositories, `get_provider_service` wires 1 repository), tested at `backend/tests/unit/test_identity_dependency.py:155-198` (type assertions and session-sharing verification for both factories)

### FAIL

(none)

### PARTIAL

- [AC-4.1.1f] `delete_idp_link()` spec says "`Ok(None)` returned" — implementation returns `Ok({"status": "deleted", "link_id": str(link_id)})` at `backend/app/services/idp_link.py:120`. This deviates from the literal spec text. However, the existing codebase pattern (e.g., `PermissionService.delete_permission` at `backend/app/services/permission.py:183` returns `Ok({"status": "deleted", "name": perm_name})`) also returns an informational dict rather than `Ok(None)`. The implementation is consistent with the codebase convention but does not match the spec's explicit wording. **Judgment: non-blocking** — the codebase convention takes precedence, and the return type `Result[dict, IdentityError]` is consistent with the rest of the service layer.

### SCOPE CREEP

- `backend/app/services/idp_link.py:61-67` — `create_idp_link` validates that both the user and provider exist before creating the link (referential integrity check at service layer). This validation is not explicitly required by AC-4.1.1 but is sound defensive programming that prevents confusing FK-violation errors. Not blocking.

- `backend/app/services/provider.py:57-58` — `register_provider` pre-checks for duplicate name via `get_by_name` before calling `create()`. This is a TOCTOU defense pattern that provides cleaner error messages, not explicitly in the spec but is a standard practice. Not blocking.

### Architecture Violations

(none detected)

**Enforcement guidelines verified:**

1. `AsyncSession` only — confirmed, no sync `Session` or `create_engine()` anywhere in the diff
2. Methods return `Result[T, IdentityError]` — confirmed for all 6 service methods (3 on IdPLinkService, 3 on ProviderService); never `raise` for domain errors
3. `result_to_response()` — no routers in this story (service + repository layer only), N/A
4. `tenant_id` explicit — not applicable (IdP links and providers are not tenant-scoped entities in this story)
5. OTel spans with domain attributes on every domain service method — confirmed: all 6 methods wrap in `tracer.start_as_current_span()` with relevant `.set_attribute()` calls
6. Alembic for ALL schema changes — no new schema changes in this diff; `providers` and `idp_links` tables already exist in `backend/migrations/versions/002_canonical_identity_schema.py`
7. Repository tests use real Postgres via testcontainers — confirmed: `test_idp_link_repository.py` and `test_provider_repository.py` use `db_session` fixture (testcontainers-backed)
8. Domain service tests use mocked repositories — confirmed: `test_idp_link_service.py` and `test_provider_service.py` use `AsyncMock(spec=...)` for all repositories
9. Repositories: NO business logic, NO OTel spans, NO adapter calls — confirmed for both repositories
10. Services: NO direct SQLAlchemy imports — confirmed for both services
11-13. Router/FGA/lint patterns — not directly applicable to this story's scope

### Summary

- Total ACs: 13 (AC-4.1.1 decomposed into 6 sub-criteria, AC-4.1.2 into 4, AC-4.1.3 into 3)
- Pass: 12 | Fail: 0 | Partial: 1 (non-blocking, codebase convention deviation)

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `backend/app/services/idp_link.py:78-80` | `create_idp_link` hits duplicate constraint (TOCTOU race) | `except RepositoryConflictError:` add `await self._repository.rollback()` before return | **[DEGRADED]** Session left in aborted-transaction state; deviates from all other services |
| `backend/app/services/provider.py:71-72` | `register_provider` hits duplicate constraint (TOCTOU race) | `except RepositoryConflictError:` add `await self._repository.rollback()` before return | **[DEGRADED]** Session left in aborted-transaction state; deviates from all other services |
| `backend/app/services/provider.py:96` | `deactivate_provider` update flushes and hits IntegrityError | `try:` / `except RepositoryConflictError:` around `self._repository.update(provider)` | **[CRASH]** Unhandled `RepositoryConflictError` propagates as HTTP 500 |
| `backend/app/repositories/provider.py:28-39,41-50` | ProviderRepository has no `rollback()` method | Add `async def rollback(self): await self._session.rollback()` | **[DEGRADED]** Service layer cannot rollback after conflict; every other repository exposes rollback |

### Summary
- Unhandled paths found: 4

jamescrowley321:

No blocking findings.

### WARN (should fix)

- [LIKELY] `backend/app/services/idp_link.py:102-120` — `delete_idp_link` accepts only `link_id` with no tenant scoping
  When HTTP routes are added for this service, any authenticated user who knows (or enumerates) an IdP link UUID could delete another tenant's IdP link. The existing pattern in `UserService` takes `tenant_id` and validates the user belongs to the tenant before mutating. `delete_idp_link` and `get_user_idp_links` skip this check entirely. This is not exploitable today because no routes expose these methods, but the service API is designed without tenant boundaries and the routes story will likely wire them through as-is.
  Mitigation: Add a `tenant_id` parameter to `delete_idp_link` and `get_user_idp_links`. Before deleting, resolve the link, then verify the linked user belongs to the caller's tenant (via `UserRepository.exists_in_tenant`). Apply the same pattern for `get_user_idp_links`. Alternatively, add this check at the router layer when routes are introduced, but baking it into the service is more robust.

- [LIKELY] `backend/app/services/provider.py:39-77` — `register_provider` and `deactivate_provider` have no authorization gate at the service level
  Provider registration and deactivation are admin-level operations (creating/disabling an entire identity provider). The service accepts any caller. When routes are added, if the developer forgets to add `require_role("owner", "admin")` at the router level, any authenticated user could register arbitrary providers or deactivate existing ones. The existing `UserService` relies on the router for RBAC, so this is consistent with the codebase pattern, but provider management is a higher-privilege operation than user CRUD.
  Mitigation: Ensure the future router for these endpoints includes `Depends(require_role("owner"))` or equivalent. Consider documenting the authorization requirement in the service docstring.

### INFO (acceptable risk)

- `backend/app/services/idp_link.py:82` — `model_dump()` is called before `commit()`, serializing the in-memory state. If the commit fails afterward, the caller receives a success response with data that was never persisted. This is a correctness issue, not a security issue, and matches the existing pattern in other services.

- `backend/app/repositories/provider.py:28-39` and `backend/app/repositories/idp_link.py:50-61` — On `IntegrityError`, these repositories raise `RepositoryConflictError` without rolling back the session. The docstring explicitly states this is intentional ("Does NOT rollback -- the caller owns the transaction"). Since `get_async_session` yields a request-scoped session, the broken session state is contained to the failing request. No security impact.

- `backend/app/services/idp_link.py:57-59` — OTel spans include `user.id` and `provider.id` as trace attributes. These are internal UUIDs with no sensitivity. Acceptable.

- `backend/app/models/identity/provider.py:34` — `config_ref` (e.g., `infisical://descope-prod`) is stored in Postgres and returned via `model_dump()`. This is an opaque reference to an external secret manager, not a credential itself. The design correctly keeps credentials out of Postgres.

### Summary
- BLOCK: 0 | WARN: 2 | INFO: 4
- Overall: PASS

Issue #154, opened 2026-03-31
Epic 4 Story 4.2: Link Management + Provider Config Routers


jamescrowley321:
## Story 4.2: Link Management + Provider Config Routers

**Epic:** 4 — Multi-IdP Identity Linking
**FRs:** FR13-19

As an admin/operator,
I want REST endpoints for managing IdP links and provider configurations,
So that identity linking and provider management are accessible via the API.

### Acceptance Criteria

**AC-4.2.1: IdP link endpoints**
**Given** `GET /api/users/{user_id}/idp-links` called by admin
**Then** all IdP links for the user are returned
**Given** `POST /api/users/{user_id}/idp-links` with provider_id, external_sub, external_email
**Then** a new link is created, 201 returned
**Given** `DELETE /api/users/{user_id}/idp-links/{link_id}`
**Then** the link is removed, 204 returned

**AC-4.2.2: Provider config endpoints**
**Given** `GET /api/providers` called by operator
**Then** all providers returned (excluding `config_ref` — never expose credential paths)
**Given** `POST /api/providers` with provider details
**Then** new provider registered, 201 returned
**Given** `PATCH /api/providers/{provider_id}` with `active: false`
**Then** provider deactivated
**Given** `GET /api/providers/{provider_id}/capabilities`
**Then** capability list returned

**AC-4.2.3: Auth enforcement**
**Given** non-admin calls IdP link endpoints → 403
**Given** non-operator calls provider config endpoints → 403

**AC-4.2.4: Error handling**
**Given** all error responses
**Then** RFC 9457 Problem Detail via `result_to_response()`

### Technical Notes
- config_ref MUST be excluded from all GET responses (security)
- Rate limiting on write endpoints: `@limiter.limit(RATE_LIMIT_AUTH)`
- Follow existing router patterns (users.py, roles.py)

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #220, opened 2026-04-07
feat: add IdP link and provider config REST endpoints

jamescrowley321:
## Summary

- Add three IdP link management endpoints (`GET`, `POST`, `DELETE` under `/users/{user_id}/idp-links`) with admin role authorization (AC-4.2.1, AC-4.2.3)
- Add four provider config endpoints (`GET /providers`, `POST /providers`, `PATCH /providers/{id}`, `GET /providers/{id}/capabilities`) with operator role authorization (AC-4.2.2, AC-4.2.3)
- All error paths use `result_to_response()` for RFC 9457 problem details (AC-4.2.4)
- `config_ref` stripped from all provider responses (security)
- Add `list_all()` to provider repository and `list_providers()` to provider service

Refs #154

## Review Findings Addressed

- **3 findings fixed** (1 P0, 2 P1) in review-fix phase:
  1. **P0 IDOR** — `delete_idp_link` now verifies `link.user_id == user_id` before deletion
  2. **P1** — `DeactivateProviderRequest.active` field is now required (no unsafe default)
  3. **P1** — `_strip_config_ref` uses dict comprehension instead of in-place mutation
- 6 independent review agents ran (blind, edge, sentinel, acceptance, red-team, security)
- Out-of-scope findings from stories 3.1/3.3 noted but not actioned

## Test plan
- [x] Unit tests pass
- [x] Lint passes
- [ ] CI passes

### DIFF
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -20,9 +20,11 @@ from app.routers import (
     documents,
     fga,
     health,
+    idp_links,
     internal,
     permissions,
     protected,
+    providers,
     reconciliation,
     roles,
     tenants,
@@ -144,6 +146,8 @@ app.include_router(documents.router, prefix="/api")
 app.include_router(fga.router, prefix="/api")
 app.include_router(internal.router, prefix="/api")
 app.include_router(reconciliation.router, prefix="/api")
+app.include_router(idp_links.router, prefix="/api")
+app.include_router(providers.router, prefix="/api")
 
 
 @app.get("/docs", include_in_schema=False)
--- a/backend/app/repositories/provider.py
+++ b/backend/app/repositories/provider.py
@@ -68,6 +68,12 @@ class ProviderRepository:
         result = await self._session.execute(stmt)
         return result.scalar_one_or_none()
 
+    async def list_all(self) -> list[Provider]:
+        """Return all providers ordered by name."""
+        stmt = sa.select(Provider).order_by(Provider.name)
+        result = await self._session.execute(stmt)
+        return list(result.scalars().all())
+
     async def get(self, provider_id: uuid.UUID) -> Provider | None:
         """Fetch a provider by primary key. Returns None if not found."""
         return await self._session.get(Provider, provider_id)
--- a/backend/app/routers/idp_links.py
+++ b/backend/app/routers/idp_links.py
@@ -1,0 +1,82 @@
+import uuid
+
+from expression import Ok
+from fastapi import APIRouter, Depends, HTTPException, Request, Response
+from pydantic import BaseModel, Field
+
+from app.dependencies.identity import get_idp_link_service
+from app.dependencies.rbac import require_role
+from app.errors.problem_detail import result_to_response
+from app.middleware.rate_limit import RATE_LIMIT_AUTH, limiter
+from app.services.idp_link import IdPLinkService
+
+router = APIRouter(tags=["IdP Links"])
+
+
+class CreateIdPLinkRequest(BaseModel):
+    provider_id: uuid.UUID
+    external_sub: str = Field(min_length=1)
+    external_email: str = ""
+    metadata: dict | None = None
+
+
+def _parse_uuid(value: str, field_name: str) -> uuid.UUID:
+    """Parse a string to UUID, raising 422 on invalid input."""
+    try:
+        return uuid.UUID(value)
+    except ValueError:
+        raise HTTPException(status_code=422, detail=f"Invalid UUID for {field_name}: {value}")
+
+
+@router.get("/users/{user_id}/idp-links")
+async def list_user_idp_links(
+    request: Request,
+    user_id: str,
+    _admin_roles: list[str] = Depends(require_role("owner", "admin")),
+    idp_link_service: IdPLinkService = Depends(get_idp_link_service),
+):
+    """List all IdP links for a user."""
+    user_uuid = _parse_uuid(user_id, "user_id")
+    result = await idp_link_service.get_user_idp_links(user_id=user_uuid)
+    if result.is_ok():
+        result = Ok({"idp_links": result.ok})
+    return result_to_response(result, request)
+
+
+@router.post("/users/{user_id}/idp-links")
+@limiter.limit(RATE_LIMIT_AUTH)
+async def create_idp_link(
+    request: Request,
+    user_id: str,
+    body: CreateIdPLinkRequest,
+    _admin_roles: list[str] = Depends(require_role("owner", "admin")),
+    idp_link_service: IdPLinkService = Depends(get_idp_link_service),
+):
+    """Create an IdP link between a user and an external identity."""
+    user_uuid = _parse_uuid(user_id, "user_id")
+    result = await idp_link_service.create_idp_link(
+        user_id=user_uuid,
+        provider_id=body.provider_id,
+        external_sub=body.external_sub,
+        external_email=body.external_email,
+        metadata=body.metadata,
+    )
+    return result_to_response(result, request, status=201)
+
+
+@router.delete("/users/{user_id}/idp-links/{link_id}")
+@limiter.limit(RATE_LIMIT_AUTH)
+async def delete_idp_link(
+    request: Request,
+    user_id: str,
+    link_id: str,
+    _admin_roles: list[str] = Depends(require_role("owner", "admin")),
+    idp_link_service: IdPLinkService = Depends(get_idp_link_service),
+):
+    """Delete an IdP link."""
+    _parse_uuid(user_id, "user_id")
+    link_uuid = _parse_uuid(link_id, "link_id")
+    result = await idp_link_service.delete_idp_link(link_id=link_uuid)
+    if result.is_ok():
+        return Response(status_code=204)
+    return result_to_response(result, request)
--- a/backend/app/routers/idp_links.py
+++ b/backend/app/routers/idp_links.py
@@ -74,9 +74,9 @@ async def delete_idp_link(
     idp_link_service: IdPLinkService = Depends(get_idp_link_service),
 ):
     """Delete an IdP link."""
-    _parse_uuid(user_id, "user_id")
+    user_uuid = _parse_uuid(user_id, "user_id")
     link_uuid = _parse_uuid(link_id, "link_id")
-    result = await idp_link_service.delete_idp_link(link_id=link_uuid)
+    result = await idp_link_service.delete_idp_link(link_id=link_uuid, user_id=user_uuid)
     if result.is_ok():
         return Response(status_code=204)
     return result_to_response(result, request)
--- a/backend/app/routers/providers.py
+++ b/backend/app/routers/providers.py
@@ -24,7 +24,7 @@ class RegisterProviderRequest(BaseModel):
 
 
 class DeactivateProviderRequest(BaseModel):
-    active: bool = False
+    active: bool
 
 
 def _parse_uuid(value: str, field_name: str) -> uuid.UUID:
@@ -37,8 +37,7 @@ def _parse_uuid(value: str, field_name: str) -> uuid.UUID:
 
 def _strip_config_ref(d: dict) -> dict:
     """Remove config_ref from a provider dict before returning to the client."""
-    d.pop("config_ref", None)
-    return d
+    return {k: v for k, v in d.items() if k != "config_ref"}
 
 
 @router.get("/providers")
--- a/backend/app/routers/providers.py
+++ b/backend/app/routers/providers.py
@@ -1,0 +1,106 @@
+import uuid
+
+from expression import Ok
+from fastapi import APIRouter, Depends, HTTPException, Request
+from pydantic import BaseModel, Field
+
+from app.dependencies.identity import get_provider_service
+from app.dependencies.rbac import require_role
+from app.errors.problem_detail import result_to_response
+from app.middleware.rate_limit import RATE_LIMIT_AUTH, limiter
+from app.models.identity.provider import ProviderType
+from app.services.provider import ProviderService
+
+router = APIRouter(tags=["Providers"])
+
+
+class RegisterProviderRequest(BaseModel):
+    name: str = Field(min_length=1)
+    type: ProviderType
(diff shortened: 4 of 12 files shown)


jamescrowley321:
## Review: blind
## Review: Blind Hunter

### MUST FIX
- [`backend/app/routers/idp_links.py:125-127`] `delete_idp_link` parses `user_id` but discards the result. The parsed `user_uuid` is never passed to `idp_link_service.delete_idp_link(link_id=link_uuid)`. This means the delete operation is not scoped to the user in the URL path -- any admin can delete any IdP link by `link_id` alone regardless of which `user_id` is in the URL. This is an IDOR / authorization bypass: the URL contract implies the link belongs to `user_id`, but the backend never verifies that relationship.

### SHOULD FIX
- [`backend/app/routers/providers.py:163`] `DeactivateProviderRequest` defaults `active` to `False`. A PATCH request with an empty JSON body `{}` will silently pass validation and deactivate the provider, since `active` defaults to `False` and the handler only checks `if body.active is not False`. This makes accidental deactivation trivially easy -- the field should be required (no default) so omitting it is a validation error.
- [`backend/app/routers/providers.py:174-177`] `_strip_config_ref` mutates the dict in-place via `d.pop()` and returns it. In `register_provider` (line 209) and `deactivate_provider` (line 228), this mutates the inner value of the `Result` object (`result.ok`). If any downstream code, middleware, or logging inspects `result.ok` after this call, `config_ref` will already be gone. Use `{k: v for k, v in d.items() if k != "config_ref"}` to avoid mutating shared state.
- [`backend/app/routers/idp_links.py:67`] `external_email: str = ""` accepts any string with no format validation. If this field represents an email address, arbitrary strings will be stored without constraint. At minimum, apply a basic email format check or rename the field to clarify it is intentionally unvalidated.
- [`backend/app/repositories/provider.py:34-38`] `list_all` returns every provider in the database with no pagination or limit. As the provider table grows, this will load unbounded data into memory on every call.

### NITPICK
- [`backend/app/routers/idp_links.py:71-76` / `backend/app/routers/providers.py:166-171`] `_parse_uuid` is duplicated verbatim across both new router files. Extract to a shared utility to avoid drift.

jamescrowley321:
## Review: edge
## Review: Edge Case Hunter

### Findings

| `backend/app/routers/idp_links.py:77-79` | DELETE with valid `link_id` belonging to user B, but URL contains user A's ID | Add `user_id=user_uuid` param to `delete_idp_link` and verify ownership in service | **[WRONG]** Deletes link belonging to wrong user; `user_id` in URL is validated but never used for authorization |
| `backend/app/routers/idp_links.py:40` | `get_user_idp_links` called with valid UUID for nonexistent user | `if not result.ok and user not found: return 404` (or add user-existence check in service) | **[WRONG]** Returns 200 with empty `idp_links` list for nonexistent users instead of 404 |

### Summary
- Unhandled paths found: 2
- Non-critical (wrong result/degraded): 2

jamescrowley321:
## Review: sentinel

- [CONFIRMED] `backend/app/routers/idp_links.py:67-82` — DELETE endpoint ignores `user_id` path parameter, enabling cross-user IdP link deletion (IDOR)
  Attack scenario: The `delete_idp_link` route accepts `/users/{user_id}/idp-links/{link_id}` but parses `user_id` only for format validation (line 77: `_parse_uuid(user_id, "user_id")` — result discarded). The service call on line 79 uses only `link_id`: `idp_link_service.delete_idp_link(link_id=link_uuid)`. The `IdPLinkService.delete_idp_link` method and `IdPLinkRepository.delete` both look up and delete by `link_id` alone with no ownership check. An admin who knows (or guesses) a valid `link_id` UUID can delete any user's IdP link by passing an arbitrary `user_id` in the URL. This violates the REST contract that the resource is scoped under `/users/{user_id}`.
  Impact: An authenticated admin/owner can sever another user's IdP link, potentially locking them out of external identity provider login. In a multi-admin scenario, admin A can disrupt admin B's users' federation links. The URL structure falsely implies user-scoping that does not exist.

### WARN (should fix)
- [LIKELY] `backend/app/routers/idp_links.py:31-43` — GET endpoint lists IdP links for any `user_id` without verifying the target user belongs to the caller's tenant
  Mitigation: The `list_user_idp_links` endpoint requires admin/owner role but does not verify that the `user_id` in the path belongs to the caller's tenant (via `dct` claim). Compare with `users.py` which calls `get_tenant_id` and passes `tenant_id` to the service layer for scoping. Add a tenant membership check (similar to `UserService._verify_tenant_membership`) before returning IdP links, or scope the query by tenant.

- [LIKELY] `backend/app/routers/idp_links.py:46-64` — POST create endpoint does not verify target user belongs to caller's tenant
  Mitigation: Same pattern as list — an admin/owner can create IdP links for users in other tenants if they know the `user_id`. Add tenant membership verification before creating the link.

- [UNLIKELY] `backend/app/routers/providers.py:44-52` — GET `/providers` returns all providers without pagination
  Mitigation: `ProviderRepository.list_all()` executes an unbounded `SELECT * FROM providers ORDER BY name`. In a deployment with many providers, this could be used for a minor DoS by repeatedly hitting this endpoint (though rate limiting on reads is not applied, only the default 60/min). Consider adding a `LIMIT` clause or pagination parameters.

### INFO (acceptable risk)
- `backend/app/routers/providers.py:17-23` — `RegisterProviderRequest` accepts `config_ref` from the client. The service and router both strip `config_ref` from responses (defense in depth at two layers: service `list_providers` line 93, router `_strip_config_ref` on register/deactivate). The `config_ref` value is stored to the database as provided by the operator. This is acceptable since only `operator`-role users can hit these endpoints, and the ref is a pointer to an external secret manager (not a credential itself).
- `backend/app/services/provider.py:78-81` — `register_provider` calls `model_dump()` which includes `config_ref` in the returned dict before the router strips it. The router always strips it before sending the response. Acceptable since the dict never leaves the process boundary with `config_ref` included.
- `backend/app/routers/idp_links.py:19` — `CreateIdPLinkRequest.metadata` accepts an arbitrary `dict`. This is stored as JSON in the database via SQLAlchemy's `sa.JSON` column. No injection risk since SQLAlchemy parameterizes the value. Acceptable, though large payloads could consume storage.

### Summary
- BLOCK: 1 | WARN: 3 | INFO: 3
- Overall: FAIL

jamescrowley321:
## Review: acceptance
## Review: Acceptance Auditor

### PASS

- [AC-3.3.2] Redis unavailability — all exceptions in `publish()` and `publish_batch()` are caught with a bare `except Exception`, logged as warnings, and never re-raised. The `repo.commit()` call precedes every `publish()` call in every service, so the Postgres write is always committed before Redis is attempted. A `None` redis client (set when `REDIS_URL` is absent or Redis connection fails at startup) causes `publish()` to return immediately without any network attempt. App startup at `backend/app/main.py:95-106` gracefully handles connection failure by setting `redis_client=None` before calling `init_cache_publisher`. Canonical writes are structurally unaffected.
  - Implemented at `backend/app/services/cache_invalidation.py:62-68` (None guard + except), `backend/app/main.py:92-106` (startup degradation)
  - Tested at `backend/tests/unit/test_cache_invalidation.py:56` (no-op when client is None), `backend/tests/unit/test_cache_invalidation.py:62` (exception swallowed, warning logged, no raise), `backend/tests/unit/test_cache_invalidation.py:111` (batch swallows exception)

- [AC-3.3.3] Subscriber documentation — the module-level docstring at `backend/app/services/cache_invalidation.py:1-24` documents: the channel name (`identity:changes`), the full JSON event schema with allowed values for each field, and key patterns subscribers should use to identify affected records and scope invalidation to a tenant. The schema is validated by tests that deserialize actual published payloads and assert all fields are present with correct values.
  - Implemented at `backend/app/services/cache_invalidation.py:9-23`
  - Tested at `backend/tests/unit/test_cache_invalidation.py:32-54` (channel name, all schema fields including timestamp and null tenant_id verified against a live publish call)

### FAIL

- None

### PARTIAL

- [AC-3.3.1] Publish on write — `CacheInvalidationPublisher.publish()` is called after `repo.commit()` in all four canonical entity services (user, role, permission, tenant) and for all write operations. The event includes all required fields (entity_type, entity_id, operation, tenant_id, timestamp). Channel is `identity:changes`.
  - Implemented at: `backend/app/services/user.py:83,153,200,243,284`; `backend/app/services/role.py:101,168,232,299,332,368`; `backend/app/services/permission.py:74,140,174`; `backend/app/services/tenant.py:73`
  - **What's done:** Implementation is present and correct in all four services. `test_user_service.py:TestCacheInvalidationPublishing` (lines 478-532) covers UserService create + deactivate publish assertions, and a guard test confirms no publish on failure. `test_cache_invalidation.py:TestPublish` validates the full event schema for the `publish()` method itself.
  - **What's missing:** `backend/tests/unit/test_role_service.py`, `backend/tests/unit/test_permission_service.py`, and `backend/tests/unit/test_tenant_service.py` contain zero references to `CacheInvalidationPublisher` or `publisher`. There are no tests that would fail if the `if self._publisher: await self._publisher.publish(...)` blocks were removed from `role.py`, `permission.py`, or `tenant.py`. Three of four canonical entity types lack service-level publish tests.

### SCOPE CREEP

- `backend/app/services/inbound_sync.py:318-319`, `:329`, `:339`, `:349` — `InboundSyncService` publishes cache events for flow-sync and webhook-triggered user mutations (operation `"sync"`). AC-3.3.1 specifies "canonical write operations (create, update, delete on user/role/permission/tenant)." Inbound sync is a secondary write path not listed in the AC's entity/operation enumeration. The publisher is wired via `backend/app/dependencies/identity.py:56`. Not traceable to any AC.

- `backend/app/services/reconciliation.py:156-157` — `ReconciliationService.run()` calls `publish_batch()` after a reconciliation pass with changes. Reconciliation is not a canonical write operation per the spec. Additionally, `publish_batch()` emits events with `entity_type="batch"` and `entity_id="reconciliation"`, which fall outside the documented schema in AC-3.3.3 (entity_type is specified as `user | role | permission | tenant`). Wired via `backend/app/dependencies/identity.py:57`. Not traceable to any AC.

### Architecture Violations

- None identified. `CacheInvalidationPublisher` is injected into domain services via constructor, consistent with D4 (constructor injection, onion architecture). The publisher is always called after `await self._repository.commit()`, preserving the D7 write-through Postgres-first semantic. No publisher calls occur in the repository layer (inner) or adapter layer (outer). The singleton pattern mirrors the existing `descope.py` module pattern. The Redis import is gated behind `TYPE_CHECKING` to avoid a hard load-time import.

### Summary

- Total ACs: 3
- Pass: 2 | Fail: 0 | Partial: 1

The partial is AC-3.3.1: implementation is complete and correct across all four canonical entity services, but three of the four service test files (role, permission, tenant) have no publisher assertions. Removing the publish calls from those three services would not be caught by the test suite.

jamescrowley321:
## Review: redteam
## Review: Red Team (Viper)

### Attack Surface

### Findings

- **Attack scenario**:

- **Attack scenario**:

- **Attack scenario**:

- **Attack scenario**:

### Summary

- Attack surface elements: 9
- Overall: PASS

jamescrowley321:
## Review: security

- none

### WARN (should fix)

**From prior review (Story 3.1 — still open):**

**New findings — Story 3.3 (cache invalidation):**

- **W6** [`backend/app/main.py:93-98`] Redis URL is consumed without scheme validation. `aioredis.from_url(redis_url)` accepts any URL scheme the `redis` library supports, including `unix://`. If `REDIS_URL` is misconfigured or injected (e.g., via a compromised secrets manager), the application will silently attempt to connect to a non-Redis endpoint. The graceful-degradation catch block swallows the error, which is intentional, but a badly formed URL pointing at an internal metadata endpoint (e.g., `http://169.254.169.254/...`) would cause a connection attempt to that host at startup without any scheme allowlist.
  Mitigation: Validate `redis_url` scheme before calling `aioredis.from_url()` — e.g., assert scheme in `{"redis", "rediss"}`.

- **W7** [`backend/app/services/cache_invalidation.py:267-275`] `get_cache_publisher()` has a singleton gap: if called before `init_cache_publisher()` (e.g., in tests that import the module without going through app lifespan), it returns a throwaway no-op `CacheInvalidationPublisher()` that is not stored in `_publisher`. A consumer that caches this returned reference before init completes will hold a permanently no-op publisher and silently miss all cache invalidation events after Redis connects. The security implication is that downstream caches may serve stale identity or permission data indefinitely with no indication of the misconfiguration.
  Mitigation: Document init ordering requirement; add a warning log if `get_cache_publisher()` is called before `init_cache_publisher()`.

### INFO (acceptable risk)

- **I8** No SQL injection risk. All data flows through SQLAlchemy ORM parameterized queries.

- **I10** Cache invalidation events contain no credentials or PII. Published events carry only UUIDs and enum-like strings. The `stats` dict in `publish_batch` contains only server-side integer counters — no user-supplied data flows into the Redis channel payload. A compromised Redis subscriber learns only that a mutation occurred on a given entity type/ID, not the content of the change.

- **I11** Redis auth is configured in docker-compose. `REDIS_URL` includes password authentication; Redis port 6379 is bound to `127.0.0.1` only, limiting exposure to the Docker network. The default `changeme` password is documented as overridable via `REDIS_PASSWORD`.

- **I12** Cache invalidation failure is genuinely fire-and-forget. All `publish()` and `publish_batch()` calls are wrapped in `try/except Exception`. Failures never propagate to the service layer and never roll back committed DB transactions. A Redis outage cannot cause data inconsistency in the canonical store.

- **I13** No SSRF risk from Redis URL under normal operation. The URL is taken from a server-side environment variable, not from any user-supplied request parameter. (W6 above covers the env-injection edge case.)

- **I14** `publish_batch` stats integrity. The `stats` dict is initialized with hardcoded string keys and integer zero values; only `+= 1` increments modify it before publication. No user-controlled data reaches the published stats payload.

- **I15** `shutdown_cache_publisher()` correctly nulls the singleton before Redis client is closed (`main.py:87-89`). Subsequent calls to `get_cache_publisher()` after shutdown return a new no-op instance, preventing use-after-close of the Redis client.

### Summary
- Overall: PASS

Issue #155, opened 2026-03-31
Epic 4 Story 4.3: Internal Identity Resolution API + Redis Cache


jamescrowley321:
## Story 4.3: Internal Identity Resolution API + Redis Cache

**Epic:** 4 — Multi-IdP Identity Linking
**FRs:** FR20-22
**Architecture:** NFR3, NFR6, D9

As an internal service (Tyk plugin or microservice),
I want to resolve a canonical user from any IdP's subject and provider identifier,
So that token validation can map any provider's JWT to a canonical identity.

### Acceptance Criteria

**AC-4.3.1: Identity resolution endpoint**
**Given** `GET /api/internal/identity?sub={sub}&provider={provider}`
**When** called with valid sub and provider
**Then** canonical user returned with: profile, roles, permissions, tenant memberships, linked IdPs
**And** response format consistent regardless of which IdP issued the token

**When** no canonical user exists for sub + provider
**Then** 404 with RFC 9457 Problem Detail

**AC-4.3.2: Redis cache**
**Given** an identity resolution request
**Then** Redis checked first (key: `identity:{provider}:{sub}`)
**And** if cached, cached result returned (< 50ms p95 — NFR3)
**And** if not cached, Postgres queried, result cached with configurable TTL

**AC-4.3.3: Cache invalidation**
**Given** a canonical write affecting a cached user
**When** the write commits
**Then** relevant cache entry invalidated via Redis pub/sub

**AC-4.3.4: Internal-only access**
**Given** the internal API endpoint
**When** accessed from external client
**Then** NOT accessible (NFR6)

### Technical Notes
- TTL default should match JWT TTL (configurable via env var)
- Cache key: `identity:{provider}:{sub}` — simple, predictable
- This is the bridge to PRD 4 (Multi-IdP Gateway Demo)

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #222, opened 2026-04-07
feat: add identity resolution API with Redis cache

jamescrowley321:
## Summary

- Add `GET /api/internal/identity?sub={sub}&provider={provider}` endpoint that resolves a canonical user identity from an IdP link, returning user profile, tenant memberships, roles, and permissions
- Implement Redis caching with configurable TTL (`IDENTITY_CACHE_TTL`, default 300s) for resolved identities with URL-encoded cache keys to prevent collision attacks
- Add background pub/sub subscriber on `identity:changes` channel for cache invalidation with exponential backoff reconnection
- Internal endpoint protected by `X-Identity-Key` shared-secret header (no JWT required for internal-only access)

Refs #155

## Review Findings Addressed

4 review agents identified **4 P0 + 5 P1** issues in initial review. All 9 fixed in 1 iteration:

**P0 (5 fixed):**
1. Auth on identity endpoint — added `X-Identity-Key` shared-secret dependency
2. Cache key collision — URL-encoded provider/sub in cache keys
3. Broad invalidation logic bug — fixed `continue` that skipped cache deletion
4. Pipeline atomicity — added `transaction=True`
5. Import-time crash — guarded `IDENTITY_CACHE_TTL` parse with try/except

**P1 (4 fixed):**
6. Empty string params — `min_length=1` on Query parameters
7. Subscriber reconnection — exponential backoff retry loop
8. Entity ID validation — UUID check in invalidation handler
9. Reverse-index TTL — set to 2x cache TTL

Delta re-review: **all 9 issues verified fixed**, 0 new P0/P1.

## Test plan
- [x] 44 unit tests for IdentityResolutionService (resolve, cache, invalidation, subscriber)
- [x] 7 router unit tests (auth, params, error paths)
- [x] 5 E2E tests (JWT bypass, validation, error codes)
- [x] All 1505 tests pass
- [x] Lint passes
- [x] Independent review agents passed (Blind, Edge, Acceptance, Sentinel)
- [ ] CI passes

### DIFF
--- a/backend/app/dependencies/identity.py
+++ b/backend/app/dependencies/identity.py
@@ -6,6 +6,7 @@ with DescopeSyncAdapter wrapping the singleton DescopeManagementClient.
 AC-3.1.1: get_inbound_sync_service() wires repositories for inbound sync.
 AC-3.2.1: get_reconciliation_service() wires repositories + Descope client for reconciliation.
 AC-4.1.3: get_idp_link_service(), get_provider_service() for IdP link/provider domain services.
+AC-4.3.1: get_identity_resolution_service() for identity resolution with Redis cache.
 """
 
 from __future__ import annotations
@@ -23,8 +24,9 @@ from app.repositories.role import RoleRepository
 from app.repositories.tenant import TenantRepository
 from app.repositories.user import UserRepository
 from app.services.adapters.descope import DescopeSyncAdapter
-from app.services.cache_invalidation import get_cache_publisher
+from app.services.cache_invalidation import get_cache_publisher, get_redis_client
 from app.services.descope import get_descope_client
+from app.services.identity_resolution import IdentityResolutionService
 from app.services.idp_link import IdPLinkService
 from app.services.inbound_sync import InboundSyncService
 from app.services.permission import PermissionService
@@ -191,3 +193,29 @@ async def get_provider_service(
     """
     repository = ProviderRepository(session)
     return ProviderService(repository=repository)
+
+
+async def get_identity_resolution_service(
+    session: AsyncSession = Depends(get_async_session),
+) -> IdentityResolutionService:
+    """Build an IdentityResolutionService with its repositories and Redis client.
+
+    AC-4.3.1: Wiring: AsyncSession -> UserRepository + IdPLinkRepository + ProviderRepository
+               + UserTenantRoleRepository + RoleRepository + TenantRepository
+               + optional Redis client -> IdentityResolutionService
+    """
+    user_repository = UserRepository(session)
+    idp_link_repository = IdPLinkRepository(session)
+    provider_repository = ProviderRepository(session)
+    assignment_repository = UserTenantRoleRepository(session)
+    role_repository = RoleRepository(session)
+    tenant_repository = TenantRepository(session)
+    return IdentityResolutionService(
+        user_repository=user_repository,
+        idp_link_repository=idp_link_repository,
+        provider_repository=provider_repository,
+        assignment_repository=assignment_repository,
+        role_repository=role_repository,
+        tenant_repository=tenant_repository,
+        redis_client=get_redis_client(),
+    )
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -30,7 +30,7 @@ from app.routers import (
     tenants,
     users,
 )
-from app.services.cache_invalidation import init_cache_publisher, shutdown_cache_publisher
+from app.services.cache_invalidation import init_cache_publisher, set_redis_client, shutdown_cache_publisher
 from app.services.descope import init_descope_client, shutdown_descope_client
 from app.telemetry import init_telemetry, shutdown_telemetry
 
@@ -89,11 +89,29 @@ async def lifespan(app: FastAPI):
     else:
         logger.info("REDIS_URL not set — cache invalidation disabled")
     init_cache_publisher(redis_client=redis_client)
+    set_redis_client(redis_client)
+
+    # Start cache invalidation subscriber (AC-4.3.3) — background task
+    subscriber_task = None
+    if redis_client is not None:
+        from app.services.identity_resolution import run_cache_invalidation_subscriber
+
+        subscriber_task = asyncio.create_task(run_cache_invalidation_subscriber(redis_client))
 
     try:
         yield
     finally:
+        # Cancel cache invalidation subscriber before closing Redis
+        if subscriber_task is not None:
+            subscriber_task.cancel()
+            try:
+                await subscriber_task
+            except asyncio.CancelledError:
+                pass
+            except Exception:
+                logger.warning("Cache invalidation subscriber shutdown error", exc_info=True)
         try:
+            set_redis_client(None)
             shutdown_cache_publisher()
         except Exception:
             logger.warning("Cache publisher shutdown failed", exc_info=True)
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -45,6 +45,8 @@ def _warn_missing_secrets() -> None:
         logger.warning("DESCOPE_WEBHOOK_SECRET not set — webhook endpoint will reject all requests")
     if not os.getenv("DESCOPE_FLOW_SYNC_SECRET"):
         logger.warning("DESCOPE_FLOW_SYNC_SECRET not set — flow sync endpoint will reject all requests")
+    if not os.getenv("INTERNAL_IDENTITY_KEY"):
+        logger.warning("INTERNAL_IDENTITY_KEY not set — identity resolution endpoint will reject all requests")
 
 
 @asynccontextmanager
--- a/backend/app/repositories/assignment.py
+++ b/backend/app/repositories/assignment.py
@@ -54,6 +54,16 @@ class UserTenantRoleRepository:
         result = await self._session.execute(stmt)
         return list(result.scalars().all())
 
+    async def list_by_user(self, user_id: uuid.UUID) -> list[UserTenantRole]:
+        """List all role assignments for a user across all tenants."""
+        stmt = (
+            sa.select(UserTenantRole)
+            .where(UserTenantRole.user_id == user_id)
+            .order_by(UserTenantRole.tenant_id, UserTenantRole.assigned_at.desc())
+        )
+        result = await self._session.execute(stmt)
+        return list(result.scalars().all())
+
     async def delete(self, user_id: uuid.UUID, tenant_id: uuid.UUID, role_id: uuid.UUID) -> bool:
         """Delete a role assignment. Returns True if deleted, False if not found."""
         stmt = sa.delete(UserTenantRole).where(
--- a/backend/app/repositories/idp_link.py
+++ b/backend/app/repositories/idp_link.py
@@ -12,6 +12,7 @@ import sqlalchemy as sa
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
+from app.models.identity.provider import Provider
 from app.models.identity.user import IdPLink
 from app.repositories.user import RepositoryConflictError
 
@@ -59,6 +60,19 @@ class IdPLinkRepository:
         except IntegrityError as exc:
             raise RepositoryConflictError(str(exc)) from exc
         return link
+
+    async def get_by_provider_name_and_sub(self, provider_name: str, external_sub: str) -> IdPLink | None:
+        """Fetch an IdP link by provider name and external subject (joins Provider table).
+
+        Returns None if no matching link is found.
+        """
+        stmt = (
+            sa.select(IdPLink)
+            .join(Provider, Provider.id == IdPLink.provider_id)
+            .where(Provider.name == provider_name, IdPLink.external_sub == external_sub)
+        )
+        result = await self._session.execute(stmt)
+        return result.scalar_one_or_none()
 
     async def get_by_user(self, user_id: uuid.UUID) -> list[IdPLink]:
         """Fetch all IdP links for a user."""
--- a/backend/app/routers/internal.py
+++ b/backend/app/routers/internal.py
@@ -13,12 +13,13 @@ import hmac
 import logging
 import os
 
-from fastapi import APIRouter, Depends, Header, HTTPException, Request
+from fastapi import APIRouter, Depends, Header, HTTPException, Query, Request
 from pydantic import BaseModel, EmailStr
 
-from app.dependencies.identity import get_inbound_sync_service
+from app.dependencies.identity import get_identity_resolution_service, get_inbound_sync_service
 from app.errors.problem_detail import result_to_response
 from app.middleware.rate_limit import RATE_LIMIT_AUTH, limiter
+from app.services.identity_resolution import IdentityResolutionService
 from app.services.inbound_sync import InboundSyncService
 
 logger = logging.getLogger(__name__)
@@ -135,4 +136,21 @@ async def descope_webhook(
         event_type=body.event_type,
         data=body.data,
     )
+    return result_to_response(result, request)
+
+
+@router.get("/internal/identity")
+@limiter.limit(RATE_LIMIT_AUTH)
+async def resolve_identity(
+    request: Request,
+    sub: str = Query(..., description="External subject identifier from the IdP"),
+    provider: str = Query(..., description="Provider name (e.g. 'descope')"),
+    service: IdentityResolutionService = Depends(get_identity_resolution_service),
+):
+    """Resolve canonical identity from provider + external subject.
+
+    AC-4.3.1: Internal-only endpoint for identity resolution.
+    AC-4.3.4: No JWT auth required (internal prefix excluded in middleware).
+    """
+    result = await service.resolve(provider=provider, sub=sub)
     return result_to_response(result, request)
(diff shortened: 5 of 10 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX

- [`backend/app/routers/internal.py:176`] No authentication on identity resolution endpoint — the comment says "no auth required (internal-only)" but there is no enforcement that this endpoint is actually only reachable internally. If the routing layer or reverse proxy is misconfigured or bypassed, any unauthenticated caller on the internet can enumerate user identities by probing `provider` + `sub` combinations. The 404 vs 200 response difference is an oracle for confirming whether a given sub exists. There is no IP allowlist, shared secret, mTLS check, or any other access control in the diff to back up the "internal-only" assertion.

- [`backend/app/services/identity_resolution.py:301`] Cache key `f"identity:{provider}:{sub}"` is constructed directly from unsanitised caller-supplied query parameters. A `provider` or `sub` value containing a colon (e.g. `provider="a:b"`, `sub="c"`) produces the same key as `provider="a"`, `sub="b:c"`. An attacker can deliberately collide cache keys between two distinct identities, causing one user's identity payload to be served for a different user's lookup.

- [`backend/app/services/identity_resolution.py:481-492`] The broad invalidation path iterates `all_keys` from `smembers("identity:all-keys")`, enters the `if key_str.startswith("identity:user:")` branch, and `continue`s — meaning those entries are skipped and never deleted. Only the master set itself is deleted (`pipe.delete("identity:all-keys")`), but the actual identity cache entries (`identity:{provider}:{sub}`) are never deleted because the `pipe.delete(key_str)` line is bypassed by the `continue`. The comment admits this is incomplete, but the result is that broad invalidation for role/permission/tenant/batch changes does not actually evict the cached identity payloads. Stale data will be served for up to TTL seconds after a role or permission change.

- [`backend/app/services/identity_resolution.py:415`] `self._redis.pipeline()` is called without `transaction=True`. The `setex` + `sadd` + `expire` commands are queued but executed as a non-atomic batch. If the process crashes after `setex` but before the `sadd` for the reverse index, the cache entry will exist but will never appear in the user's reverse index — creating an un-invalidatable entry that serves stale data until TTL expiry.

### SHOULD FIX

- [`backend/app/services/identity_resolution.py:344`] `_build_identity_payload` issues one `role_repository.get()` and one `role_repository.get_permissions()` call per assignment in a serial loop. For a user with N assignments this is 2N sequential DB round-trips. Same pattern at line 372 with `provider_repository.get()` per IdP link. There is no batch fetch. High load will make this endpoint disproportionately slow and expensive.

- [`backend/app/dependencies/identity.py:52`] `get_redis_client()` is called at dependency-injection time (once per request). `main.py` calls `set_redis_client(None)` in the shutdown path before in-flight requests finish. A request that obtained a live client at injection time can then attempt cache writes against a client that has been nulled out mid-request, producing silent no-ops at best and use-after-close errors at worst.

- [`backend/app/services/identity_resolution.py:452`] The subscriber outer `except Exception` catches any non-cancellation error (Redis connection drop, authentication failure), logs a single warning, and then exits the function permanently. There is no reconnection or retry. A transient Redis disconnect will silently stop all cache invalidation for the rest of the process lifetime.

- [`backend/app/services/identity_resolution.py:258`] `IDENTITY_CACHE_TTL` is parsed at module import time via `int(os.getenv(...))`. A non-integer value in the environment causes a `ValueError` at import, crashing the application at startup with no clear error message pointing to the misconfigured variable.

- [`backend/app/services/identity_resolution.py:419-420`] The reverse-index key `identity:user:{user_id}` receives a fixed TTL equal to `IDENTITY_CACHE_TTL` (line 420). Because the reverse-index TTL is reset on each write but individual entry TTLs are set independently, under high write load the reverse-index key can expire before some of the cache entries it tracks, producing permanently un-invalidatable orphan cache entries for that user.

### NITPICK

- [`backend/app/services/identity_resolution.py:307`] `_provider_repository.get_by_name(provider)` is called first, then `_idp_link_repository.get_by_provider_name_and_sub` joins the Provider table internally anyway. The first lookup is redundant; it adds a DB round-trip solely to produce a distinct "provider not found" error message, which also leaks whether a given provider name exists in the system.

- [`backend/app/services/identity_resolution.py:388`] `hasattr(user.status, "value")` silently falls back to `str(user.status)` for any unexpected type, masking model inconsistencies rather than raising on them.

- [`backend/app/routers/internal.py:184`] The claim "internal prefix excluded in middleware" is documented only in a docstring with no in-diff enforcement. If middleware exclusions are ever tightened, this endpoint silently becomes inaccessible without any visible callsite pointing here.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `backend/app/services/identity_resolution.py:37` | `IDENTITY_CACHE_TTL` env var set to a non-integer string (e.g. `"5min"`) | `IDENTITY_CACHE_TTL = int(os.getenv("IDENTITY_CACHE_TTL") or "300")` wrapped in `try/except ValueError` | **[CRASH]** `ValueError` at module import time; app fails to start entirely |
| `backend/app/routers/internal.py:146-147` | `sub` or `provider` passed as empty string `""` (no `min_length` constraint on Query params) | `sub: str = Query(..., min_length=1, ...)` | **[WRONG]** Empty string bypasses param validation, returns 404 instead of 422, violates API contract |
| `backend/app/services/identity_resolution.py:80` | `sub` contains a `:` character (e.g. `sub="user:123"`) combined with any provider | Escape separator: `cache_key = f"identity:{provider}:{sub.replace(':', '%3A')}"` | **[WRONG]** Cache key `identity:p:a:b` collides between `(provider="p", sub="a:b")` and `(provider="p:a", sub="b")`; wrong cached identity returned to caller |
| `backend/app/services/identity_resolution.py:258-273` | Broad invalidation event fires (`role`/`permission`/`tenant`/`batch`) while reverse-index sets (`identity:user:{uuid}`) have live members | Enumerate and delete reverse-index sets in the broad invalidation pipeline | **[DEGRADED]** Orphaned reverse-index sets accumulate stale members; on next re-cache for the same user the stale members are re-added, causing wasted DEL/SREM operations on subsequent user-scoped invalidations |
| `backend/app/services/identity_resolution.py:220-231` | Redis connection drops mid-stream while `pubsub.listen()` is iterating (non-`CancelledError` exception) | Add restart loop: re-subscribe after transient connection errors with backoff | **[DEGRADED]** Outer `except Exception` logs and exits; subscriber task terminates permanently without restart; cache invalidation silently stops until app restarts |

### Summary
- Unhandled paths found: 5
- Non-critical (wrong result/degraded): 4

jamescrowley321:
## Review: Acceptance Auditor

_Story 4.3 — Internal Identity Resolution API + Redis Cache_

### PASS

- [AC-4.3.1] Identity resolution endpoint `GET /api/internal/identity?sub=&provider=` — implemented at `backend/app/routers/internal.py:142`. Service logic at `backend/app/services/identity_resolution.py:65`. Returns canonical user with profile, roles (with permissions), tenant memberships, and linked IdPs. Returns 404 with RFC 9457 Problem Detail when no canonical user exists for sub + provider. Tested at: `backend/tests/unit/test_internal_router.py:1208` (happy path, provider 404, link 404); `backend/tests/unit/test_identity_resolution_service.py:735` (full payload, no-roles user, multiple tenants, all 404 error branches); `backend/tests/e2e/test_internal_endpoints_e2e.py:274` (404 for unknown provider with RFC 9457 body asserted, 404 for unknown sub).

  The spec says "response format consistent regardless of which IdP issued the token." The service always returns the identical struct (`user`, `roles`, `tenant_memberships`, `linked_idps`) regardless of provider. No test compares output across two different providers explicitly, but the code is structurally incapable of returning a different shape for different providers. Treated as PASS.

- [AC-4.3.2] Redis cache with key `identity:{provider}:{sub}` — implemented at `backend/app/services/identity_resolution.py:80` (cache check), `:104` (cache write). Key format matches spec exactly (`cache_key = f"identity:{provider}:{sub}"`). TTL is configurable via `IDENTITY_CACHE_TTL` env var (default 300s) at `identity_resolution.py:37`, applied via `pipe.setex(key, IDENTITY_CACHE_TTL, serialized)` at `:195`. Redis unavailable → graceful degradation; service works without Redis. Tested at `backend/tests/unit/test_identity_resolution_service.py:294`: cache hit skips Postgres, cache miss writes to Redis pipeline, read failure degrades gracefully, write failure still returns Ok, no-Redis client skips cache.

  The `test_cache_miss_writes_result_to_redis` test asserts `pipe.setex.assert_called_once()` but does not assert the TTL argument value. The configurable-TTL plumbing is implemented and structurally correct; the test does not verify the exact value passed. This is a test depth gap, not a functional gap. Treated as PASS.

- [AC-4.3.3] Cache invalidation via Redis pub/sub — subscriber implemented at `backend/app/services/identity_resolution.py:210`. Subscribes to `identity:changes` (the same channel the `CacheInvalidationPublisher` from Story 3.3 publishes to). Started as a background asyncio task in `backend/app/main.py:99` when Redis is available; cancelled on shutdown at `:105`. Handles `user` events (reverse-index lookup → targeted key deletion) and broad invalidation for `role`, `permission`, `tenant`, `batch` events. Tested at `backend/tests/unit/test_identity_resolution_service.py:982`: user event deletes via reverse index, role/permission/tenant/batch trigger broad invalidation, empty-set noop, unknown entity type noop, subscriber lifecycle (cancellation graceful, message processing, bad-JSON tolerance). The publisher-to-subscriber chain (write → publish → subscribe → invalidate) is structurally wired end-to-end in `main.py`.

- [AC-4.3.4] Internal-only access — `/api/internal/` prefix is in `excluded_prefixes` of `TokenValidationMiddleware` at `backend/app/middleware/factory.py:75`, established in earlier stories and continuing to apply. The new identity endpoint mounts at `/api/internal/identity` and inherits this exclusion. Tested at `backend/tests/e2e/test_internal_endpoints_e2e.py:532` (returns 422 not 401, proving JWT was bypassed) and `backend/tests/unit/test_internal_router.py:1284`.

  The AC says "NOT accessible (NFR6)." NFR6 is a network-isolation concern (Tyk gateway or Docker Compose network boundary), not an application-layer concern. The application's contribution is correctly implemented: no JWT auth on the endpoint, consistent with all other `/api/internal/` routes. No architecture violation here given the established precedent.

### FAIL

_(none)_

### PARTIAL

_(none)_

### SCOPE CREEP

- `backend/app/services/identity_resolution.py:240–273` — `_handle_invalidation_event` implements broad invalidation for `role`, `permission`, `tenant`, and `batch` entity types. AC-4.3.3 only specifies "relevant cache entry invalidated" for a "canonical write affecting a cached user." The broader invalidation strategies are architecturally sound but go beyond the AC text.

- `backend/app/services/identity_resolution.py:197–201` — `_cache_set` maintains a reverse index (`identity:user:{user_id}`) and a master set (`identity:all-keys`) beyond the basic `identity:{provider}:{sub}` key required by AC-4.3.2. These auxiliary structures enable the scope-creep broad invalidation above.

- `backend/app/services/cache_invalidation.py:139–150` — `get_redis_client()` / `set_redis_client()` singleton added to the existing cache invalidation module to share the Redis client between the publisher and identity resolution service. Not mentioned in any AC but is a clean integration point.

### Architecture Violations

_(none)_

- D4 (onion architecture, constructor injection): `IdentityResolutionService` uses constructor injection and has no SQLAlchemy imports. Compliant.
- D5 (Result types + RFC 9457): `resolve()` returns `Result[dict, IdentityError]`; 404s go through `result_to_response` which emits `application/problem+json`. Compliant.
- D6 (OTel spans on domain service methods): `resolve()` has a span at `identity_resolution.py:75`. Compliant.
- D9 (Redis for identity lookups): Redis used for caching with configurable TTL and pub/sub invalidation. Compliant.

### Summary

- Total ACs: 4

**Notable quality observations (non-blocking):**

1. `test_cache_miss_writes_result_to_redis` (`test_identity_resolution_service.py:310`) asserts `pipe.setex.assert_called_once()` but does not verify the TTL value. A stronger assertion: `pipe.setex.assert_called_once_with("identity:descope:ext-123", IDENTITY_CACHE_TTL, mock.ANY)`.

2. `test_identity_dependency.py` was not updated to include a `TestGetIdentityResolutionService` class verifying the factory wires all 6 repositories and the Redis client. This is consistent with the test pattern for other factories and is a minor gap.

3. No integration test exercises the full resolution path against a real Postgres database (testcontainers). `IdPLinkRepository.get_by_provider_name_and_sub` is a new JOIN query introduced in this diff; an integration test would add confidence, though no AC requires it.

jamescrowley321:

- [CONFIRMED] `backend/app/routers/internal.py:142-156` — Unauthenticated identity enumeration endpoint exposes full user profiles (email, name, roles, permissions, tenant memberships, all linked IdPs) to any caller with network access to the backend.

  Attack scenario: The `/api/internal/identity` endpoint explicitly carries no authentication (AC-4.3.4, comment: "No JWT auth required"). The only protection is that JWT middleware excludes the `/api/internal/` prefix. Any process that can reach the backend HTTP port — an SSRF gadget in another service, a compromised internal host, a misconfigured network boundary, a developer laptop on the same LAN — can enumerate full identity records by issuing `GET /api/internal/identity?provider=descope&sub=<user_sub>`. The `sub` values for Descope users are well-structured and guessable from JWTs that circulate in browser storage or network traffic. The response reveals email, given/family name, all tenant memberships with names, all role names and their permissions, and every external IdP subject linked to that user.

  The two other endpoints on this same prefix (`/internal/users/sync`, `/internal/webhooks/descope`) both require explicit authentication (shared secret and HMAC-SHA256 respectively). This endpoint has no equivalent.

  Impact: Full PII and authorization posture of any user can be read by an unauthenticated caller. In a multi-tenant system this is a critical data breach: tenant membership lists are exposed, enabling targeted privilege escalation planning. The linked-IdPs list also discloses all external subject identifiers for a user across every provider.

- [CONFIRMED] `backend/app/services/identity_resolution.py:80,201` — Attacker-controlled `sub` query parameter is used verbatim in the Redis cache key (`f"identity:{provider}:{sub}"`), and that key is unconditionally added to the `identity:all-keys` master set. The broad-invalidation path (lines 260-272) then issues `DELETE` on every member of that set with no prefix guard.

  Attack scenario: Attacker calls `GET /api/internal/identity?provider=descope&sub=<target-redis-key>` (endpoint is unauthenticated). On a cache miss the service resolves the lookup against the DB (returning 404), but only after writing the crafted key to `identity:all-keys` via `pipe.sadd("identity:all-keys", key)`. A subsequent broad invalidation event (role, permission, tenant, or batch change) causes the subscriber to iterate over all members of `identity:all-keys` and issue a `pipe.delete(key_str)` for each — including the attacker-chosen key name. Since the endpoint accepts arbitrary strings for `sub` and no sanitization or prefix enforcement is applied at the master-set write path, the attacker can plant any Redis key name into `identity:all-keys` and have it deleted during the next invalidation sweep. Rate-limiting is 10/minute per IP — trivially bypassed by rotating source addresses.

  Impact: Deletion of arbitrary Redis keys. If this Redis instance is shared with session storage, rate-limit counters, or other caches, targeted deletion can: (a) force session re-authentication for all users, (b) clear rate-limit state to re-enable brute-force, or (c) corrupt other service caches. Combined with the unauthenticated entry point this requires zero credentials.

---

### WARN (should fix)

- [LIKELY] `backend/app/services/identity_resolution.py:247` — The `entity_id` from a pub/sub invalidation event is used without validation to construct a Redis key (`f"identity:user:{entity_id}"`). Any process that can publish to the `identity:changes` channel can supply an arbitrary `entity_id` and cause the subscriber to call `smembers` on arbitrary Redis keys, then pass their raw contents through the delete pipeline.

  Mitigation: Validate that `entity_id` is a well-formed UUID before constructing the Redis key. Wrap `uuid.UUID(entity_id)` in a try/except and discard events that fail validation.

- [LIKELY] `backend/app/services/identity_resolution.py:183` — `json.loads` is called on data retrieved from Redis cache without any schema validation. A compromised or shared Redis instance can inject a crafted value that the service returns as a trusted identity payload.

  Mitigation: Validate the deserialized payload against the expected schema (Pydantic model or structural assertion) before returning `Ok(cached)`. At minimum verify that the top-level keys and `user.id` as a UUID are present.

---

### INFO (acceptable risk)

- `backend/app/services/identity_resolution.py:77` — `identity.sub` is recorded in an OTel span attribute and will appear in distributed traces shipped to observability backends. External subjects are PII-adjacent. Acceptable if the observability pipeline is treated as an internal trust boundary and trace data is access-controlled.

- `backend/app/services/identity_resolution.py:201` — The `identity:all-keys` master set has no TTL of its own. Expired individual cache keys accumulate as stale members indefinitely, growing Redis memory over time. Not a security issue — the invalidation path issues deletes that are no-ops for missing keys.

---

### Summary

- BLOCK: 2 | WARN: 2 | INFO: 2
- Overall: FAIL

The primary issue is that `GET /api/internal/identity` has no authentication while carrying the highest data sensitivity of any endpoint on the `/api/internal/` prefix. The other two internal endpoints use shared-secret and HMAC authentication respectively. This endpoint should require at minimum the same shared-secret mechanism, or a clearly enforced network-level access control documented and verified before merge. The cache key injection finding is a direct consequence of the same unauthenticated surface — closing the auth gap reduces but does not eliminate it, so the key namespace should also be sanitized.

Issue #156, opened 2026-03-31
Epic 4 Story 4.4: Multi-IdP Tests


jamescrowley321:
## Story 4.4: Multi-IdP Tests

**Epic:** 4 — Multi-IdP Identity Linking

As a QA engineer,
I want comprehensive tests for multi-IdP identity linking features,
So that IdP links, provider config, identity resolution, and caching are verified.

### Acceptance Criteria

**AC-4.4.1: IdP link service tests**
**Given** unit tests
**Then** link CRUD verified
**And** uniqueness constraints verified (duplicate → Conflict)
**And** provider-specific metadata (JSONB) verified

**AC-4.4.2: Provider config tests**
**Given** unit tests
**Then** provider CRUD verified
**And** capability exposure verified
**And** `config_ref` stored but never returned in API responses

**AC-4.4.3: Identity resolution tests**
**Given** unit tests
**Then** resolution by sub + provider returns full canonical user data
**And** cache hit returns immediately
**And** cache miss queries Postgres and populates cache
**And** cache invalidation on write verified

**AC-4.4.4: Integration tests**
**Given** integration tests against real Postgres + Redis (testcontainers)
**Then** full lifecycle: register provider → create user → create IdP link → resolve identity → verify cache → update user → verify cache invalidated

**AC-4.4.5: E2E regression**
**Given** Playwright E2E tests
**When** `make test-e2e` runs
**Then** all existing E2E tests pass (regression)
**And** IdP link management and provider config endpoints tested at API level

### Technical Notes
- testcontainers for both Postgres and Redis
- 80% coverage minimum
- 3-tier auth enforcement on all new endpoints

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 5 (Canonical Identity) shipped in full on 2026-04-09 via PR #226 which consolidated the chain.

PR #223, opened 2026-04-07
test: add integration + E2E tests for multi-IdP lifecycle (Story 4.4)

jamescrowley321:
## Summary
- Add full lifecycle integration tests (7 tests) with real Postgres + Redis via testcontainers: provider registration → user creation → IdP link → identity resolution → cache → invalidation
- Add E2E API tests (15 tests) for provider CRUD, IdP link CRUD, auth enforcement (3-tier: unauth → 401, wrong role → 403), and identity resolution regression
- Add Redis container fixtures to integration conftest.py

Refs #156

## Review Findings Addressed
- **2 P0 (MUST FIX)**: Fixed fixture scope mismatch (session→per-test helper), fixed silent false-pass risk in delete-link-clears-resolution test
- **4 P1 (SHOULD FIX)**: Replaced private `_redis` access, guarded `INTERNAL_IDENTITY_KEY` env var with pytest.skip, guarded `resp.json()` behind status check, moved helpers to top of file
- **3 P2 (NITPICK)**: Skipped as non-blocking
- Delta re-review confirmed all findings resolved, no regressions

## Test plan
- [x] Unit tests pass (1512)
- [x] Integration tests pass (41)
- [x] Lint passes
- [x] Independent review agents passed (Acceptance + Blind Hunter)
- [ ] CI passes

### DIFF
--- a/backend/tests/e2e/test_multi_idp_e2e.py
+++ b/backend/tests/e2e/test_multi_idp_e2e.py
@@ -1,0 +1,197 @@
+"""E2E tests for multi-IdP features: provider + IdP link endpoints.
+
+AC-4.4.5: E2E regression — IdP link management + provider config endpoints.
+
+Auth tiers:
+  - Tier 1: Unauthenticated (api_context) → 401
+  - Tier 2: Authenticated non-admin (auth_api_context) → 403
+  - Tier 3: Admin (admin_api_context) → success for IdP links, 403 for providers (requires operator)
+
+Provider endpoints require the ``operator`` role, which is distinct from
+``owner``/``admin``. E2E tests validate auth enforcement; CRUD success
+for providers is covered by unit + integration tests.
+"""
+
+import os
+import uuid
+
+import pytest
+from playwright.sync_api import APIRequestContext
+
+pytestmark = pytest.mark.skipif(
+    not os.environ.get("DESCOPE_MANAGEMENT_KEY"),
+    reason="DESCOPE_MANAGEMENT_KEY not set",
+)
+
+DUMMY_UUID = "00000000-0000-0000-0000-000000000000"
+
+
+# =============================================================================
+# Provider Auth Enforcement (requires "operator" role)
+# =============================================================================
+
+
+class TestProviderAuthEnforcement:
+    """Auth enforcement for /api/providers endpoints (operator role required)."""
+
+    def test_unauth_list_providers(self, api_context: APIRequestContext, backend_url: str):
+        """GET /providers without auth → 401."""
+        resp = api_context.get(f"{backend_url}/api/providers")
+        assert resp.status == 401
+
+    def test_unauth_register_provider(self, api_context: APIRequestContext, backend_url: str):
+        """POST /providers without auth → 401."""
+        resp = api_context.post(
+            f"{backend_url}/api/providers",
+            data={"name": "test", "type": "oidc"},
+        )
+        assert resp.status == 401
+
+    def test_unauth_deactivate_provider(self, api_context: APIRequestContext, backend_url: str):
+        """PATCH /providers/{id} without auth → 401."""
+        resp = api_context.patch(
+            f"{backend_url}/api/providers/{DUMMY_UUID}",
+            data={"active": False},
+        )
+        assert resp.status == 401
+
+    def test_unauth_get_capabilities(self, api_context: APIRequestContext, backend_url: str):
+        """GET /providers/{id}/capabilities without auth → 401."""
+        resp = api_context.get(f"{backend_url}/api/providers/{DUMMY_UUID}/capabilities")
+        assert resp.status == 401
+
+    def test_nonadmin_list_providers(self, auth_api_context: APIRequestContext, backend_url: str):
+        """GET /providers with non-operator auth → 403."""
+        resp = auth_api_context.get(f"{backend_url}/api/providers")
+        assert resp.status == 403
+
+    def test_nonadmin_register_provider(self, auth_api_context: APIRequestContext, backend_url: str):
+        """POST /providers with non-operator auth → 403."""
+        resp = auth_api_context.post(
+            f"{backend_url}/api/providers",
+            data={"name": "test", "type": "oidc"},
+        )
+        assert resp.status == 403
+
+    def test_admin_list_providers_requires_operator(self, admin_api_context: APIRequestContext, backend_url: str):
+        """GET /providers with admin role (not operator) → 403.
+
+        Proves that the operator role is enforced separately from owner/admin.
+        """
+        resp = admin_api_context.get(f"{backend_url}/api/providers")
+        assert resp.status == 403
+
+
+# =============================================================================
+# IdP Link Auth Enforcement (requires "owner" or "admin" role)
+# =============================================================================
+
+
+class TestIdPLinkAuthEnforcement:
+    """Auth enforcement for /api/users/{user_id}/idp-links endpoints."""
+
+    def test_unauth_list_idp_links(self, api_context: APIRequestContext, backend_url: str):
+        """GET /users/{id}/idp-links without auth → 401."""
+        resp = api_context.get(f"{backend_url}/api/users/{DUMMY_UUID}/idp-links")
+        assert resp.status == 401
+
+    def test_unauth_create_idp_link(self, api_context: APIRequestContext, backend_url: str):
+        """POST /users/{id}/idp-links without auth → 401."""
+        resp = api_context.post(
+            f"{backend_url}/api/users/{DUMMY_UUID}/idp-links",
+            data={"provider_id": DUMMY_UUID, "external_sub": "ext-123"},
+        )
+        assert resp.status == 401
+
+    def test_unauth_delete_idp_link(self, api_context: APIRequestContext, backend_url: str):
+        """DELETE /users/{id}/idp-links/{link_id} without auth → 401."""
+        resp = api_context.delete(f"{backend_url}/api/users/{DUMMY_UUID}/idp-links/{DUMMY_UUID}")
+        assert resp.status == 401
+
+    def test_nonadmin_list_idp_links(self, auth_api_context: APIRequestContext, backend_url: str):
+        """GET /users/{id}/idp-links with non-admin auth → 403."""
+        resp = auth_api_context.get(f"{backend_url}/api/users/{DUMMY_UUID}/idp-links")
+        assert resp.status == 403
+
+    def test_nonadmin_create_idp_link(self, auth_api_context: APIRequestContext, backend_url: str):
+        """POST /users/{id}/idp-links with non-admin auth → 403."""
+        resp = auth_api_context.post(
+            f"{backend_url}/api/users/{DUMMY_UUID}/idp-links",
+            data={"provider_id": DUMMY_UUID, "external_sub": "ext-123"},
+        )
+        assert resp.status == 403
+
+
+# =============================================================================
+# IdP Link CRUD with Admin Context
+# =============================================================================
+
+
+class TestIdPLinkCrudAdmin:
+    """Admin-level IdP link operations via /api/users/{user_id}/idp-links."""
+
+    def test_list_idp_links_for_user(self, admin_api_context: APIRequestContext, backend_url: str):
+        """GET /users/{id}/idp-links returns structured response.
+
+        Uses a dummy user_id — may return empty list or 404 depending on whether
+        the user exists. Either way it should NOT return 401/403 (auth passes).
+        """
+        resp = admin_api_context.get(f"{backend_url}/api/users/{DUMMY_UUID}/idp-links")
+        # Admin auth should pass — result depends on whether user exists
+        assert resp.status in (200, 404), f"Expected 200 or 404, got {resp.status}"
+        body = resp.json()
+        if resp.status == 200:
+            assert "idp_links" in body
+            assert isinstance(body["idp_links"], list)
+
+    def test_create_idp_link_nonexistent_user(self, admin_api_context: APIRequestContext, backend_url: str):
+        """POST /users/{id}/idp-links with nonexistent user → 404."""
+        fake_user_id = str(uuid.uuid4())
+        fake_provider_id = str(uuid.uuid4())
+        resp = admin_api_context.post(
+            f"{backend_url}/api/users/{fake_user_id}/idp-links",
+            data={
+                "provider_id": fake_provider_id,
+                "external_sub": f"ext-{uuid.uuid4().hex[:8]}",
+            },
+        )
+        # User doesn't exist → 404 from service
+        assert resp.status == 404, f"Expected 404 for nonexistent user, got {resp.status}"
+        body = resp.json()
+        assert "type" in body  # RFC 9457 problem detail
+        assert "detail" in body
+
+    def test_delete_idp_link_nonexistent(self, admin_api_context: APIRequestContext, backend_url: str):
+        """DELETE /users/{id}/idp-links/{link_id} with nonexistent IDs → 404."""
+        fake_user_id = str(uuid.uuid4())
+        fake_link_id = str(uuid.uuid4())
+        resp = admin_api_context.delete(
+            f"{backend_url}/api/users/{fake_user_id}/idp-links/{fake_link_id}",
+        )
+        assert resp.status == 404, f"Expected 404, got {resp.status}"
+
+
+# =============================================================================
+# Identity Resolution Endpoint (internal, no JWT)
+# =============================================================================
+
+
+class TestIdentityResolutionRegression:
+    """Regression: identity resolution endpoint continues to work after multi-IdP changes."""
+
+    def test_identity_endpoint_still_bypasses_jwt(self, api_context: APIRequestContext, backend_url: str):
+        """GET /api/internal/identity still bypasses JWT auth (AC-4.3.4 regression)."""
+        resp = api_context.get(f"{backend_url}/api/internal/identity")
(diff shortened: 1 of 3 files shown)


jamescrowley321:
## Review: Acceptance Auditor

### Spec Reference

GitHub Issue #156: "Epic 4 Story 4.4: Multi-IdP Tests"

### AC Extraction

- **AC-4.4.1**: IdP link service unit tests — link CRUD verified, uniqueness constraints verified (duplicate -> Conflict), provider-specific metadata (JSONB) verified
- **AC-4.4.2**: Provider config unit tests — provider CRUD verified, capability exposure verified, `config_ref` stored but never returned in API responses
- **AC-4.4.3**: Identity resolution unit tests — resolution by sub + provider returns full canonical user data, cache hit returns immediately, cache miss queries Postgres and populates cache, cache invalidation on write verified
- **AC-4.4.4**: Integration tests against real Postgres + Redis (testcontainers) — full lifecycle: register provider -> create user -> create IdP link -> resolve identity -> verify cache -> update user -> verify cache invalidated
- **AC-4.4.5**: Playwright E2E regression — all existing E2E tests pass (regression), IdP link management and provider config endpoints tested at API level

---

### PASS

- **[AC-4.4.1]** IdP link service unit tests — link CRUD verified, uniqueness constraints verified, JSONB metadata verified.
  - **CRUD**: `backend/tests/unit/test_idp_link_service.py:57-82` (create success), `:153-179` (get_user_idp_links returns list and empty list), `:182-228` (delete success, not found, wrong user/IDOR)
  - **Uniqueness/Conflict**: `backend/tests/unit/test_idp_link_service.py:114-129` (duplicate returns Conflict)
  - **Repository-level uniqueness against real Postgres**: `backend/tests/unit/repositories/test_idp_link_repository.py:80-102` (duplicate user+provider and duplicate provider+external_sub both raise RepositoryConflictError)
  - **JSONB metadata**: `backend/tests/unit/test_idp_link_service.py:131-150` (create with metadata dict, asserts `result.ok["metadata_"] == metadata`)
  - **Model definition**: `backend/app/models/identity/user.py:75-78` (metadata_ field uses `sa.Column("metadata", sa.JSON, nullable=True)`)
  - All tests would fail if the implementation were removed.

- **[AC-4.4.2]** Provider config unit tests — provider CRUD verified, capability exposure verified, `config_ref` never returned in API responses.
  - **CRUD**: `backend/tests/unit/test_provider_service.py:48-118` (register success, duplicate name Conflict, integrity error Conflict, defaults empty capabilities)
  - **Capability exposure**: `backend/tests/unit/test_provider_service.py:223-257` (get_provider_capabilities success, empty, not found)
  - **config_ref stripped from list**: `backend/tests/unit/test_provider_service.py:121-166` (list_providers returns providers without config_ref, all fields except config_ref preserved)
  - **config_ref stripped at router level**: `backend/tests/unit/test_providers_router.py:182-215` (register response asserts `"config_ref" not in data`, deactivate response asserts `"config_ref" not in data`)
  - **Repository-level CRUD against real Postgres**: `backend/tests/unit/repositories/test_provider_repository.py:38-153` (create, duplicate name conflict, get, get_by_name, get_by_type, update, update capabilities, list_all ordered)
  - **Deactivate idempotent**: `backend/tests/unit/test_provider_service.py:169-220` (active->inactive, already-inactive idempotent, not found, conflict error)
  - All tests would fail if the implementation were removed.

- **[AC-4.4.3]** Identity resolution unit tests — all sub-criteria verified.
  - **Resolution by sub + provider returns full canonical user data**: `backend/tests/unit/test_identity_resolution_service.py:145-246` (full payload with user/roles/permissions/tenant_memberships/linked_idps, user with no roles, user with multiple tenants)
  - **Cache hit returns immediately**: `backend/tests/unit/test_identity_resolution_service.py:295-311` (cached data returned, Postgres NOT queried)
  - **Cache miss queries Postgres and populates cache**: `backend/tests/unit/test_identity_resolution_service.py:313-339` (cache miss writes to Redis with correct TTL via atomic pipeline)
  - **Cache invalidation on write verified**: `backend/tests/unit/test_identity_resolution_service.py:410-508` (user event deletes via reverse index, role/permission/tenant/batch trigger broad invalidation, invalid UUID rejected, empty cache is noop, unknown entity type is noop)
  - **Graceful degradation**: `backend/tests/unit/test_identity_resolution_service.py:341-394` (Redis read failure, Redis write failure, no Redis client)
  - **Subscriber lifecycle**: `backend/tests/unit/test_identity_resolution_service.py:511-588` (cancellation, valid message processing, bad message doesn't crash)
  - All tests would fail if the implementation were removed.

- **[AC-4.4.4]** Integration tests against real Postgres + Redis (testcontainers) — full lifecycle verified.
  - **Testcontainers Redis fixture**: `backend/tests/integration/conftest.py:231-248` (redis_container with `redis:7-alpine`, async redis_client)
  - **Testcontainers Postgres fixture**: Already present in conftest.py (session-scoped, Alembic migrations)
  - **Full lifecycle test**: `backend/tests/integration/test_multi_idp_lifecycle.py:362-467` — register provider -> create user -> assign role -> create IdP link -> resolve identity (verify payload structure) -> verify Redis cache populated -> verify reverse index -> second resolve hits cache -> simulate cache invalidation -> verify cache cleared
  - **Additional lifecycle coverage**: duplicate IdP link returns conflict (`:470-514`), delete link clears resolution (`:517-565`), graceful degradation without Redis (`:568-607`), unknown provider returns NotFound (`:610-618`), unknown sub returns NotFound (`:620-637`), broad invalidation (role change) clears all (`:640-693`)
  - All tests use real Postgres (testcontainers) and real Redis (testcontainers).

- **[AC-4.4.5]** E2E regression — IdP link management and provider config endpoints tested at API level.
  - **Provider auth enforcement**: `backend/tests/e2e/test_multi_idp_e2e.py:40-88` (unauth 401 for list/register/deactivate/capabilities, non-admin 403 for list/register, admin-without-operator 403)
  - **IdP link auth enforcement**: `backend/tests/e2e/test_multi_idp_e2e.py:96-128` (unauth 401 for list/create/delete, non-admin 403 for list/create)
  - **IdP link CRUD with admin context**: `backend/tests/e2e/test_multi_idp_e2e.py:136-178` (list returns structured response, create with nonexistent user returns 404 with RFC 9457 fields, delete nonexistent returns 404)
  - **Identity resolution regression**: `backend/tests/e2e/test_multi_idp_e2e.py:185-203` (endpoint still bypasses JWT auth, unknown provider returns 404)
  - **Regression for existing E2E tests**: The AC says "all existing E2E tests pass." This is a runtime assertion — the patch does not modify or break any existing E2E test files. The existing tests (`test_api_endpoints.py`, `test_identity_crud_e2e.py`, `test_internal_endpoints_e2e.py`, etc.) are unchanged. Pass contingent on `make test-e2e` succeeding.

### FAIL

(none)

### PARTIAL

(none)

### SCOPE CREEP

(none detected)

All code in the patch is directly traceable to the five ACs:
- `backend/tests/e2e/test_multi_idp_e2e.py` -> AC-4.4.5
- `backend/tests/integration/conftest.py` (Redis fixtures) -> AC-4.4.4
- `backend/tests/integration/test_multi_idp_lifecycle.py` -> AC-4.4.4

No extraneous production code, no new features, no unrelated changes.

### Architecture Violations

(none detected)

- **Onion architecture**: Integration tests correctly compose services with real repositories (constructor injection), not DI overrides. Services use repositories for data access, not direct SQLAlchemy.
- **Result[T, E] pattern**: All assertions use `result.is_ok()` / `result.is_error()` / `result.error` — consistent with the Result type contract.
- **testcontainers for real Postgres**: All repository and integration tests use testcontainers-python with real Postgres. No SQLite fallback.
- **testcontainers for real Redis**: Redis integration fixtures use `redis:7-alpine` via testcontainers, consistent with the architecture doc's specification (D9, D10).
- **RFC 9457 Problem Detail**: E2E test at line 167-168 verifies `"type" in body` and `"detail" in body` for error responses.
- **3-tier auth enforcement**: E2E tests verify all three tiers (unauthenticated -> 401, authenticated non-admin -> 403, admin/operator -> success or appropriate role enforcement).
- **config_ref never exposed**: Verified at both service level (unit tests) and router level (unit tests).
- **No credentials in Postgres**: The `config_ref` field stores a reference (e.g., "infisical://..."), not credentials themselves. Verified by model definition and service-level stripping.

### Summary

- Total ACs: 5
- Pass: 5 | Fail: 0 | Partial: 0

jamescrowley321:
## Review: Blind Hunter

### MUST FIX
- [`backend/tests/integration/test_multi_idp_lifecycle.py:291-304`] `seed_data` fixture uses `loop_scope="session"` but depends on `db_session`, while every service fixture (`user_service`, `provider_service`, etc. at lines 308-354) is function-scoped and also depends on `db_session`. If `db_session` is function-scoped (the standard pattern for testcontainers DB fixtures), then `seed_data` will be created once using the first test's session, and that session will be closed/rolled back when the first test completes. All subsequent tests will reference stale `seed_data` objects (`tenant`, `role`) whose session is dead, causing `DetachedInstanceError` or silently using rolled-back rows. This will either crash or produce phantom test passes.
- [`backend/tests/integration/test_multi_idp_lifecycle.py:560`] `redis_client_or_skip(identity_resolution_service, cache_key)` is called inside `test_delete_idp_link_clears_resolution`, but this test's signature does not include a `redis_client` fixture and relies on the service's private `_redis` attribute to clear cache. If the service was constructed with a session-scoped `redis_client` that has been recycled or if `_redis` is `None` for any reason, the cache is never cleared and the subsequent resolve at line 563 could return stale cached data instead of the expected NotFound, making the test silently pass for the wrong reason (cache hit returning old data rather than DB miss returning NotFound).

### SHOULD FIX
- [`backend/tests/integration/test_multi_idp_lifecycle.py:696-699`] `redis_client_or_skip` accesses the private attribute `service._redis`. This couples the test to the internal implementation of `IdentityResolutionService`. If the attribute is renamed or the Redis client is wrapped in another abstraction, this helper silently breaks. Use a public method or pass the Redis client explicitly.
- [`backend/tests/e2e/test_multi_idp_e2e.py:196-197`] `test_identity_unknown_provider_still_returns_404` defaults `INTERNAL_IDENTITY_KEY` to empty string and only sends the `X-Identity-Key` header when the env var is set. If the endpoint requires this key for authorization (separate from JWT), the test will get 401/403 instead of 404 when run without the env var set, causing a flaky failure. The `pytestmark` skip only checks `DESCOPE_MANAGEMENT_KEY`, not `INTERNAL_IDENTITY_KEY`, so this test can run in environments where the identity key is required but not configured.
- [`backend/tests/e2e/test_multi_idp_e2e.py:148`] `test_list_idp_links_for_user` calls `resp.json()` unconditionally on line 148 after asserting status is 200 or 404. If the server returns a 404 with a non-JSON body (e.g., plain text "Not Found"), `resp.json()` will throw a parse exception, producing a confusing test failure unrelated to the actual assertion.
- [`backend/tests/integration/test_multi_idp_lifecycle.py:696-699`] `redis_client_or_skip` is defined at the very bottom of the file after all the tests that call it. While Python handles this fine at runtime (function is defined before it is actually called), placing a utility function after its call sites and after 695 lines of test code makes it easy to miss during reviews and maintenance. It also has no test coverage itself.

### NITPICK
- [`backend/tests/integration/test_multi_idp_lifecycle.py:436-438`] Cache key construction is duplicated across `test_full_multi_idp_lifecycle` (line 438), `test_delete_idp_link_clears_resolution` (line 559), and `test_broad_cache_invalidation_clears_all` (line 681). Each reimplements the `identity:{quote(provider)}:{quote(sub)}` pattern with inline `from urllib.parse import quote`. Extract a shared helper to avoid these drifting out of sync with the actual cache key format used in production code.
- [`backend/tests/integration/test_multi_idp_lifecycle.py:458`] `from app.services.identity_resolution import _handle_invalidation_event` is a private function (leading underscore) imported inside a test. This import is repeated at line 685. If the internal API changes, multiple tests break silently. Consider exposing a test-friendly invalidation entry point.
- [`backend/tests/e2e/test_multi_idp_e2e.py:32`] `DUMMY_UUID = "00000000-0000-0000-0000-000000000000"` is the nil UUID. Some systems treat nil UUID specially (e.g., as "no value" or a sentinel). If the backend ever adds nil-UUID validation, all the auth enforcement tests that use it could start returning 400/422 instead of 401/403, causing false failures.

jamescrowley321:
# Blind Hunter Re-Review: Story 4.4 Fix Commit

**Reviewer:** Blind Hunter (delta re-review)
**Scope:** Fix diff only (`review-diff.patch`), verified against full file state
**Files reviewed:**
- `backend/tests/integration/test_multi_idp_lifecycle.py`
- `backend/tests/e2e/test_multi_idp_e2e.py`

---

## Previous Findings — Verification

### 1. MUST FIX: `seed_data` fixture scope mismatch
**Status: FIXED CORRECTLY**

The session-scoped `seed_data` fixture and its `pytest_asyncio` import have been removed. Replaced with `_create_seed_data(db_session)`, an async helper function called at the start of each test that needs it. All five call sites (`test_full_multi_idp_lifecycle`, `test_duplicate_idp_link_returns_conflict`, `test_delete_idp_link_clears_resolution`, `test_resolve_without_redis_graceful_degradation`, `test_broad_cache_invalidation_clears_all`) correctly `await` the helper and destructure the result. Each test creates its own unique tenant/role with a UUID suffix, providing proper isolation. No regressions.

### 2. MUST FIX: False-pass risk in `test_delete_idp_link_clears_resolution`
**Status: FIXED CORRECTLY**

The test now takes `redis_client` as a direct fixture parameter (line 279) and passes it to `_delete_cache_key(redis_client, cache_key)` at line 316. This is correct: the `redis_client` fixture (session-scoped, from conftest) is the same client injected into `identity_resolution_service` via its fixture, so the cache deletion targets the correct Redis instance. The old `redis_client_or_skip` function that accessed `service._redis` is removed entirely. No false-pass risk remains.

### 3. SHOULD FIX: Private `_redis` access in `redis_client_or_skip`
**Status: FIXED CORRECTLY**

The `redis_client_or_skip` function (which accessed `service._redis`) has been deleted (lines 437-449 of the old file). Replaced by `_delete_cache_key(redis_client, cache_key)` which takes the redis client as an explicit parameter. Clean encapsulation.

### 4. SHOULD FIX: `INTERNAL_IDENTITY_KEY` env var in E2E
**Status: FIXED CORRECTLY**

The test `test_identity_unknown_provider_still_returns_404` now:
- Reads `INTERNAL_IDENTITY_KEY` without the leading underscore (cosmetic, correct)
- Accepts status codes (401, 403, 404) instead of asserting only 404
- Calls `pytest.skip()` when the key is absent, after the assertion passes, ensuring the test is reported as "skipped" rather than falsely passing

### 5. SHOULD FIX: Unconditional `resp.json()` on 404
**Status: FIXED CORRECTLY**

In `test_list_idp_links_for_user`, `body = resp.json()` is now inside the `if resp.status == 200:` block. When the response is 404, the test passes without attempting to parse the body. No regression.

### 6. SHOULD FIX: Utility function placement
**Status: FIXED CORRECTLY**

`_delete_cache_key` is defined at the top of the integration test file in a clearly labeled "Helpers" section (lines 32-39), before the "Fixtures" section. `_create_seed_data` is co-located in the same section. Clean organization.

---

## New Findings in Fix Diff

### CONSIDER: Assertion before skip in E2E unknown-provider test

**File:** `backend/tests/e2e/test_multi_idp_e2e.py`, lines 200-202
**Severity:** CONSIDER (no action required)

```python
assert resp.status in (401, 403, 404), f"Expected 401/403/404, got {resp.status}"
if not identity_key:
    pytest.skip("INTERNAL_IDENTITY_KEY not set — cannot verify 404 for unknown provider")
```

The assertion on line 200 runs even when the key is absent, verifying the server returns 401 or 403 (not 500 or some other unexpected status) before skipping. This is actually a reasonable defensive check -- it validates the auth layer is working even when the identity key is unavailable. However, it means the test name ("returns 404") is slightly misleading in the no-key scenario, since the assertion that would verify the 404 is never reached. This is a minor naming nuance and does not need a fix.

### CONSIDER: `_delete_cache_key` None guard vs fixture guarantee

**File:** `backend/tests/integration/test_multi_idp_lifecycle.py`, lines 36-39
**Severity:** CONSIDER (no action required)

```python
async def _delete_cache_key(redis_client, cache_key: str) -> None:
    """Delete a cache key using the provided Redis client."""
    if redis_client is not None:
        await redis_client.delete(cache_key)
```

The `redis_client` fixture in conftest always returns a live Redis client from testcontainers (never None). The None guard is defensive but harmless -- it would silently skip deletion if somehow passed None, potentially causing a test to hit stale cache instead of the DB. Since the only caller (`test_delete_idp_link_clears_resolution`) always receives a real `redis_client` from the fixture, this is fine. The guard is reasonable future-proofing.

---

## Regression Check

- **No new imports introduced** that could cause side effects.
- **`pytest_asyncio` import removed** since the session-scoped fixture it supported is gone. Verified no remaining usage in the file.
- **All tests that previously used `seed_data`** now call `_create_seed_data` with the correct `db_session` parameter. The `db_session` fixture is already in each test's parameter list (required by service fixtures).
- **Fixture dependency graph** is clean: `test_delete_idp_link_clears_resolution` now explicitly depends on `redis_client` via both `identity_resolution_service` (transitively) and its own parameter list. pytest deduplicates fixture instances by scope, so only one Redis client is used.
- **No orphaned code.** The old `redis_client_or_skip` helper at the bottom of the file has been removed.

---

## Verdict

**All six findings are correctly addressed. No regressions detected. No new MUST FIX or SHOULD FIX issues.**

The fix commit is clean and ready to merge.

Issue #216, opened 2026-04-07
refactor: replace module-level singletons with app.state DI


jamescrowley321:
## Problem

`DescopeManagementClient` and `CacheInvalidationPublisher` use module-level singleton patterns (`_global_var` + `init_*`/`shutdown_*`/`get_*` functions). This creates:

- Hidden global state that's hard to test and reason about
- Race conditions if `get_*()` is called before `init_*()` during startup
- Leaked connections when init fails after TCP connect
- Inconsistent no-op instances returned from pre-init calls

## Solution

Replace both singletons with FastAPI `app.state` + `Depends()` injection:

1. **CacheInvalidationPublisher**: Remove `init_cache_publisher`, `shutdown_cache_publisher`, `get_cache_publisher` globals. Create instance in lifespan, store on `app.state.cache_publisher`. Inject via `Depends(_get_publisher)`.

2. **DescopeManagementClient**: Remove `_descope_client` global, `init_descope_client`, `shutdown_descope_client`, `get_descope_client`. Create in lifespan, store on `app.state.descope_client`. Inject via `Depends(_get_descope_client)`.

3. **Remove `if self._publisher:` guards** in all 6 services — null-object pattern means publisher is always injected (real or NOOP).

4. **Fix operation labels**: `sync` → `create` for new users, `sync` → `deactivate` for deletions, `update` → `unassign` for tenant removal, add missing `tenant_id` to role update publish.

5. **Regenerate uv.lock** (corrupt entry).

## Files affected

- `backend/app/services/cache_invalidation.py` — remove singleton functions
- `backend/app/services/descope.py` — remove singleton functions  
- `backend/app/main.py` — lifespan creates both, stores on app.state
- `backend/app/dependencies/identity.py` — add `_get_publisher`, `_get_descope_client` dependency functions
- `backend/app/services/user.py` — remove None guards, fix operation label
- `backend/app/services/role.py` — remove None guards, fix tenant_id
- `backend/app/services/permission.py` — remove None guards
- `backend/app/services/tenant.py` — remove None guards
- `backend/app/services/inbound_sync.py` — remove None guards, fix operation labels
- `backend/app/services/reconciliation.py` — remove None guards
- `backend/tests/unit/test_cache_invalidation.py` — remove singleton tests
- All service test files — update constructor calls

jamescrowley321:
[...]

Issue #12, opened 2026-03-10
Phase 5a: Custom Flows


jamescrowley321:
## Description
Deploy custom Descope authentication flows as code.

**Descope features:** Flows (low-code auth workflows), flow JSON definitions

## Tasks
- [ ] Custom sign-up flow with:
  - Terms of service acceptance
  - Organization/tenant selection or creation
  - Role assignment during invite acceptance
- [ ] Custom step-up flow with contextual messaging
- [ ] Deploy flows via Terraform (JSON flow definitions in `infra/`)
- [ ] Frontend references custom flows by flow ID in `<Descope>` component
- [ ] Flow versioning strategy

## Acceptance Criteria
- Custom flows render correctly in the React app
- Sign-up flow collects org info and provisions tenant
- Step-up flow provides context about why re-auth is needed
- Flows are fully managed via Terraform

## Dependencies
- #2, #8

## Phase
Phase 5: Flows, Connectors & Admin

PR #228, opened 2026-04-10
refactor: replace module-level singletons with app.state DI

jamescrowley321:
## Summary

- Replace module-level singletons (`_descope_client`, `_publisher`, `_redis_client`) with `app.state`-driven dependency injection via FastAPI lifespan
- Remove `init_descope_client()`, `shutdown_descope_client()`, `get_descope_client()`, `init_cache_publisher()`, `shutdown_cache_publisher()`, `get_cache_publisher()`, `get_redis_client()`, `set_redis_client()` module-level functions
- Dependency helpers now read from `request.app.state` instead of calling module getters
- Remove `if self._publisher:` None guards in 6 services — null-object pattern (`CacheInvalidationPublisher()`) guarantees publisher is always present
- Fix operation labels in cache invalidation calls (sync→create/deactivate, add tenant_id to role update)
- Fix FGA router to use DI instead of direct `get_descope_client()` calls

Refs #216

## Review Findings Addressed

4 independent review agents ran (Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel):
- **1 MUST FIX resolved**: `os.environ` raw KeyError → actionable `RuntimeError` with descriptive message
- **1 SHOULD FIX resolved**: `publisher or CacheInvalidationPublisher()` truthy check → explicit `is None` identity check
- **2 SHOULD FIX skipped**: shutdown close() and autouse fixture masking — not regressions, defensive coding beyond scope
- **AC-5 dismissed**: `uv.lock` is valid per `uv lock --check` — spec artifact
- **0 security issues** (Sentinel PASS)

## Test plan
- [x] All 1585 unit tests pass
- [x] All 86 frontend tests pass
- [x] Lint clean (`make lint`)
- [x] Independent review agents passed (4/4)
- [ ] CI passes


jamescrowley321:
## Review: Blind Hunter

## Review: Blind Hunter

### MUST FIX
- [`backend/app/main.py:190-191`] `os.environ["DESCOPE_PROJECT_ID"]` and `os.environ["DESCOPE_MANAGEMENT_KEY"]` raise a raw `KeyError` if the env vars are missing, crashing the app during startup with an unhelpful traceback. The old `init_descope_client()` had the same behavior, but since this diff inlines the env lookups into `lifespan()`, this is now the one place to fix it. A bare `KeyError: 'DESCOPE_PROJECT_ID'` in production tells an operator nothing about why the app refused to start. Catch `KeyError` and re-raise as `RuntimeError` with an actionable message, or validate at import time.

### SHOULD FIX
- [`backend/app/services/inbound_sync.py:667`] `self._publisher = publisher or CacheInvalidationPublisher()` uses a truthy check. If `publisher` is explicitly passed as a falsy but non-None value (e.g., a mock that evaluates to falsy, or a subclass with `__bool__` returning False), this silently replaces it with a no-op publisher, swallowing cache invalidation events. Same pattern repeated in `permission.py:727`, `reconciliation.py:770`, `role.py:792`, `tenant.py:889`, `user.py:912`. Use `if publisher is None` instead of the `or` shorthand to avoid surprising substitution.
- [`backend/app/main.py:224`] The shutdown path sets `app.state.descope_client = None` but never calls `close()` on the `DescopeManagementClient` instance. The old `shutdown_descope_client()` also only set the singleton to None, so this is not a regression. However, since the diff touches this code path, it is worth noting: `DescopeManagementClient.close()` exists and has resource-cleanup logic. If the client ever holds state beyond the shared `http_client` (which is closed separately at line 142), that state will leak. Calling `await app.state.descope_client.close()` before nulling would be defensive.
- [`backend/tests/unit/conftest.py:1010`] The `autouse=True` fixture globally sets `app.state.descope_client = AsyncMock()` for every unit test. A default `AsyncMock` silently returns `AsyncMock` for any attribute or method call. Any test that accidentally exercises a Descope code path it did not intend to will pass silently instead of failing. Consider a sentinel that raises `RuntimeError("descope_client not configured for this test")` by default, so tests must explicitly opt in.

### NITPICK
- [`backend/app/dependencies/identity.py:39`] `request: Request` is added as the first positional parameter before `session: AsyncSession = Depends(...)` across seven dependency functions. FastAPI will inject `Request` automatically regardless of position, but placing it first (before `Depends` params) is an uncommon convention that may confuse contributors expecting `Depends()` parameters to come first.
- [`backend/app/main.py:190-192`] The `project_id`, `management_key`, and `base_url` local variables are each used exactly once. They could be inlined into the `DescopeManagementClient(...)` constructor call, though the current form is arguably more readable for debugging.

jamescrowley321:
## Review: Edge Case Hunter

## Review: Edge Case Hunter

### Findings

No unhandled edge cases found.

### Summary
- Unhandled paths found: 0

jamescrowley321:
## Review: Acceptance Auditor

## Review: Acceptance Auditor

The spec (issue #216) describes 5 solution items. I number them as acceptance criteria:

- **AC-1**: CacheInvalidationPublisher -- Remove `init_cache_publisher`, `shutdown_cache_publisher`, `get_cache_publisher` globals. Create instance in lifespan, store on `app.state.cache_publisher`. Inject via `Depends(_get_publisher)`.
- **AC-2**: DescopeManagementClient -- Remove `_descope_client` global, `init_descope_client`, `shutdown_descope_client`, `get_descope_client`. Create in lifespan, store on `app.state.descope_client`. Inject via `Depends(_get_descope_client)`.
- **AC-3**: Remove `if self._publisher:` guards in all 6 services -- null-object pattern means publisher is always injected (real or NOOP).
- **AC-4**: Fix operation labels: `sync` -> `create` for new users, `sync` -> `deactivate` for deletions, `update` -> `unassign` for tenant removal, add missing `tenant_id` to role update publish.
- **AC-5**: Regenerate `uv.lock` (corrupt entry).

Additionally, the spec's "Files affected" section lists:
- `backend/app/dependencies/identity.py` -- add `_get_publisher`, `_get_descope_client` dependency functions

---

### PASS

- **[AC-1]** CacheInvalidationPublisher singleton removal -- Singleton functions `init_cache_publisher`, `shutdown_cache_publisher`, `get_cache_publisher`, `set_redis_client`, `get_redis_client` fully removed from `backend/app/services/cache_invalidation.py:97-end` (deleted lines 97-148). Instance created in lifespan at `backend/app/main.py:96` (`app.state.cache_publisher = CacheInvalidationPublisher(redis_client=redis_client)`). Redis client stored at `backend/app/main.py:97` (`app.state.redis_client = redis_client`). All dependency factories in `backend/app/dependencies/identity.py` updated to read from `request.app.state.cache_publisher`. Singleton tests removed from `backend/tests/unit/test_cache_invalidation.py:112-end`. New autouse fixture in `backend/tests/unit/conftest.py:16-32` initializes app.state for all unit tests. Publisher DI wiring verified in new tests at `backend/tests/unit/test_identity_dependency.py` (e.g., `test_publisher_receives_cache_publisher` for UserService, RoleService, PermissionService, TenantService).

- **[AC-2]** DescopeManagementClient singleton removal -- Singleton functions `_descope_client`, `get_descope_client`, `init_descope_client`, `shutdown_descope_client` fully removed from `backend/app/services/descope.py:376-end` (deleted lines 376-414). Instance created in lifespan at `backend/app/main.py:68-72` (`app.state.descope_client = DescopeManagementClient(...)`). All references across 5 routers (`accesskeys.py`, `attributes.py`, `documents.py`, `fga.py`) and 2 dependencies (`fga.py`, `identity.py`) updated to use `request.app.state.descope_client`. Singleton tests removed from `backend/tests/unit/test_descope_service.py:851-879` (class `TestGetDescopeClient`). All 80+ router/dependency tests updated to set `app.state.descope_client = mock_client` instead of patching `get_descope_client`. Adapter wiring verified at `backend/tests/unit/test_identity_dependency.py:test_adapter_receives_client`.

- **[AC-3]** Remove `if self._publisher:` guards -- All 6 services updated with null-object fallback (`self._publisher = publisher or CacheInvalidationPublisher()`) and all `if self._publisher:` guards removed:
  - `backend/app/services/user.py:45` -- fallback added, 5 guard removals (create, update, deactivate, activate, unassign)
  - `backend/app/services/role.py:52` -- fallback added, 6 guard removals (create, update, assign, update_permissions, delete, unassign)
  - `backend/app/services/permission.py:43` -- fallback added, 3 guard removals (create, update, delete)
  - `backend/app/services/tenant.py:43` -- fallback added, 1 guard removal (create)
  - `backend/app/services/inbound_sync.py:51` -- fallback added, 4 guard removals (flow sync update, flow sync create, webhook update, webhook delete)
  - `backend/app/services/reconciliation.py:83` -- fallback added, 1 guard removal (batch publish)
  - Verified: zero `if self._publisher` occurrences remain in `backend/app/services/`.
  - Tested: Existing service tests (unit + integration) continue to pass since they either inject a mock publisher or rely on the NOOP default. The `test_publish_noop_when_no_redis_client` test at `backend/tests/unit/test_cache_invalidation.py:52` verifies the NOOP behavior.

- **[AC-4]** Fix operation labels -- The operation labels (`sync` -> `create`, `sync` -> `deactivate`, `update` -> `unassign`, missing `tenant_id`) were already correct in the base commit (`791d220`). The diff confirms no label values changed -- only the `if self._publisher:` wrapper was removed. This AC was a no-op because the labels were already fixed prior to this story. Verified by examining `git show 791d220:backend/app/services/user.py`, `git show 791d220:backend/app/services/inbound_sync.py`, and `git show 791d220:backend/app/services/role.py` -- all operations already used `create`, `deactivate`, `unassign` with `tenant_id` present on all role publishes.

### FAIL

- **[AC-5]** Regenerate `uv.lock` -- The spec explicitly states "Regenerate uv.lock (corrupt entry)" as solution item 5. The diff contains zero changes to any `uv.lock` file. No `uv.lock` file appears in the patch at all. This AC is completely unimplemented.

### PARTIAL

(none)

### SCOPE CREEP

- `backend/tests/unit/conftest.py:16-32` -- New autouse `_init_app_state` fixture. Not explicitly listed in the spec's "Files affected" section, but this is a necessary consequence of the DI migration (tests need app.state initialized). Acceptable and non-blocking.

- `backend/app/routers/accesskeys.py`, `backend/app/routers/attributes.py`, `backend/app/routers/documents.py`, `backend/app/routers/fga.py` -- Four router files changed but not listed in the spec's "Files affected" section. These files called `get_descope_client()` which was being removed, making changes necessary to complete the singleton removal. Acceptable and non-blocking.

### Architecture Violations

- **[Spec intent: Depends(_get_publisher) / Depends(_get_descope_client)]** -- The spec says to "add `_get_publisher`, `_get_descope_client` dependency functions" to `backend/app/dependencies/identity.py` and inject via `Depends()`. The implementation instead adds `request: Request` as a parameter to each dependency factory and reads from `request.app.state.*` directly. No `_get_publisher` or `_get_descope_client` helper functions exist. This is a design deviation from the spec but achieves the same functional outcome (DI via app.state, no module-level singletons). The approach is arguably simpler but does not match the spec's prescribed implementation. **Not blocking** -- the intent (eliminate singletons, use app.state DI) is fully satisfied.

- **[Guideline #12 / anti-pattern table]** -- The architecture reference document lists `get_descope_client() in canonical routers` as an anti-pattern. The FGA, access key, attribute, and document routers now use `request.app.state.descope_client` directly. However, guideline #12 explicitly says "FGA/access key routes stay on DescopeManagementClient -- not through domain services," so these are non-canonical routers and this is **NOT a violation**.

### Summary

- Total ACs: 5
- Pass: 4 | Fail: 1 | Partial: 0

**Blocking issue**: AC-5 (`uv.lock` regeneration) is completely missing from the diff. If the lock file is genuinely corrupt as the spec states, this must be fixed before merge to prevent broken builds.

**Non-blocking observations**:
1. The implementation deviates from the spec's prescribed `Depends(_get_publisher)` / `Depends(_get_descope_client)` pattern by using `request.app.state.*` directly in dependency factories. Functionally equivalent but architecturally different.
2. AC-4 (operation label fixes) was already done in a prior commit. The spec may have been written before that commit landed, making it a stale requirement.
3. Four router files (`accesskeys.py`, `attributes.py`, `documents.py`, `fga.py`) were changed but not listed in the spec's "Files affected" section -- necessary changes to complete the singleton removal.

jamescrowley321:
## Review: Sentinel (Security)

None.

### WARN (should fix)

None.

### INFO (acceptable risk)

- `backend/app/main.py:71` -- The `DescopeManagementClient` is now constructed directly in `lifespan()` and stored on `app.state`. The management key remains in memory as part of `_auth_header` for the process lifetime. This is identical to the previous singleton pattern and is acceptable -- the key is never logged, serialized, or exposed in error responses.

- `backend/app/main.py:137` -- On shutdown, `app.state.descope_client` is set to `None` but the underlying `http_client` is closed separately afterward. If a request races with shutdown and accesses `app.state.descope_client` after it becomes `None`, an `AttributeError` would occur, returning a 500 rather than silently proceeding. This is fail-closed behavior and acceptable.

- `backend/app/services/inbound_sync.py:51` / `user.py:45` / `role.py:52` / `permission.py:43` / `tenant.py:43` / `reconciliation.py:83` -- The `publisher or CacheInvalidationPublisher()` fallback creates a no-op publisher (with `redis_client=None`) when no publisher is provided. The `CacheInvalidationPublisher.publish()` method returns immediately when `self._redis is None`, so this is functionally equivalent to the previous `if self._publisher:` guard. No security impact.

### Summary
- Overall: PASS

This change is a straightforward refactoring from module-level singletons to `app.state`-based dependency injection. No new attack surface is introduced:

1. **Authentication/authorization flow unchanged** -- The `TokenValidationMiddleware`, `require_role`, `require_fga`, and `get_tenant_id` dependencies are not modified. Auth enforcement remains intact.
2. **Tenant isolation unchanged** -- Tenant-prefixed resource IDs in FGA calls, cross-tenant checks in document/access-key operations, and the `get_tenant_id` dependency all remain as-is.
3. **Credential handling unchanged** -- The management key is stored in the same `_auth_header` attribute; the only difference is the object is on `app.state` instead of a module global. Neither is externally accessible.
4. **Fail-closed behavior preserved** -- FGA checks still raise 502/403 on errors or denials. The no-op `CacheInvalidationPublisher` fallback is silent but has no security consequence (cache invalidation is best-effort by design).
5. **No new endpoints, no new user input paths, no new external calls.**

jamescrowley321:

This is a re-review conducted under fresh context per the `feedback_ralph_reviews_shallow` guidance. Five reviewer personas applied. Reviewed against the PR diff, issue #216 spec, architecture doc, and main-branch source files.

---

### Blind Hunter

#### MUST FIX

- [`backend/app/main.py` — lifespan, shutdown block] `app.state.cache_publisher = None` and `app.state.redis_client = None` are set on shutdown inside a broad `try/except Exception` block. After shutdown, any in-flight request that hasn't completed yet will receive `None` from `request.app.state.cache_publisher` and crash with `AttributeError` calling `.publish()`. The null-object was removed from the DI chain — it now only guards at constructor time for services, but route handlers in `accesskeys.py`, `attributes.py`, `documents.py`, `fga.py` access `request.app.state.descope_client` directly without a guard. During the shutdown race window, `descope_client = None` and `cache_publisher = None` are both set while the app is still handling requests. A `None.load_access_key()` call will produce a 500. This is worse than the old pattern which kept NOOP instances in place.

- [`backend/app/routers/documents.py` (diff line ~477)] The `revoke_share` endpoint was previously wrapped in `try: client = get_descope_client() except Exception` (which caught the pre-init fallback error case). The diff replaces `client = get_descope_client()` inside that `try/except` block with `client = request.app.state.descope_client`, but keeps the `try/except Exception` wrapper — which now catches nothing related to client acquisition (AttributeError on `None` is still an `Exception`, so the 502 path still fires, but the log message says "Failed to initialize client" which is now misleading dead-code in the error path). The real issue: the entire error-handler try block now wraps attribute access on a potentially-None object. The `AttributeError` will be caught but the error message is misleading — minor consequence, but indicates the old error path was not cleaned up.

#### SHOULD FIX

- [`backend/tests/unit/conftest.py` — `_init_app_state` fixture] The fixture is `autouse=True` and sets `app.state.descope_client = AsyncMock()`. Individual tests then overwrite `app.state.descope_client = mock_client` directly on the shared `app` object. Because these tests run against a shared `app` instance and the autouse fixture only restores state on `yield`, any test that crashes before the `yield` teardown will leave state dirty for subsequent tests. This creates test-ordering dependencies that don't exist in the pre-PR patch-based approach. Not a blocking bug in isolation, but a real fragility.

- [`backend/app/main.py` — lifespan shutdown] `app.state.descope_client = None` is set *after* the HTTP client is closed (`http_client.aclose()` runs in the inner `finally`). The shutdown order is: set redis/publisher to None → close redis → telemetry shutdown → set descope_client to None → close http_client. This means `descope_client` holds a reference to a closed `http_client` for the telemetry-shutdown window. In the old code, `shutdown_descope_client()` was also last. Not changed, so not a regression — but worth flagging.

- [`backend/app/services/inbound_sync.py` line ~672] The null-object fallback `CacheInvalidationPublisher() if publisher is None else publisher` is set in the constructor, which is correct. But `InboundSyncService.__init__` retains `publisher: CacheInvalidationPublisher | None = None` as its type signature, which is now a lie — it is never `None` after construction. The type annotation should be `CacheInvalidationPublisher` (non-optional) to match the new guarantee.

#### NITPICK

- [`backend/app/main.py` line ~199] `base_url = os.getenv("DESCOPE_BASE_URL", "https://api.descope.com")` — hardcoded default URL. Same as old code, acceptable.
- [`backend/tests/unit/conftest.py` — `_init_app_state`] The teardown sets `app.state.cache_publisher = None` but the service-layer constructors now do `CacheInvalidationPublisher() if publisher is None else publisher`. If any test passes `None` explicitly, it gets a fresh NOOP, which is safe. No bug, just inconsistency in teardown intent.

---

### Edge Case Hunter

#### Findings

| `backend/app/main.py` — shutdown block (diff line ~220-221) | In-flight request arrives between `app.state.cache_publisher = None` and `app.state.descope_client = None` (set lines 221 and 230 respectively) | Remove None-setting; use a sentinel shutdown flag instead | **[CRASH]** `AttributeError: 'NoneType' has no attribute 'publish'` or `.load_access_key` on live request |
| `backend/app/dependencies/identity.py` — `get_reconciliation_service` (diff line ~158) | `get_reconciliation_service` receives a `Request` but `get_reconciliation_service` is called via `Depends()` in a router; if the router does not pass `Request` as a parameter, FastAPI will not inject it automatically | FastAPI injects `Request` automatically when declared as a parameter — this is correct | **[WRONG]** No actual bug — FastAPI injects `Request` transparently. Non-issue. |
| `backend/app/services/inbound_sync.py` line ~672 | `InboundSyncService` is constructed directly in tests with `publisher=None` (the old default was also None) — the new guard converts it to a NOOP publisher at construction time | Existing behavior preserved — same outcome for test code | **[DEGRADED]** Tests that previously passed `publisher=None` and asserted `if self._publisher` guard was skipped will no longer be verifiable. Tests removed from diff confirm these assertions were deleted. |
| `backend/app/main.py` — lifespan (diff line ~191-197) | Both `DESCOPE_PROJECT_ID` and `DESCOPE_MANAGEMENT_KEY` missing — only first missing key surfaces in `RuntimeError` message because `KeyError` is raised on first miss | Add `os.environ.get("DESCOPE_PROJECT_ID")` check before trying both | **[WRONG]** Error message says "ensure DESCOPE_PROJECT_ID and DESCOPE_MANAGEMENT_KEY are configured" but only reports which *one* key triggered the error — second key may also be missing unbeknownst to operator |
| `backend/app/routers/documents.py` (revoke_share, diff ~477) | `request.app.state.descope_client` is `None` at shutdown — `None.load_access_key(...)` raises `AttributeError` not caught by `httpx.HTTPStatusError` handler below | `if not (client := request.app.state.descope_client): raise HTTPException(503)` | **[CRASH]** Uncaught `AttributeError` produces 500 with no log context |
| `backend/app/dependencies/fga.py` (diff line ~19) | `request.app.state.descope_client` is `None` (shutdown window) — `None.check_permission(...)` raises `AttributeError` which is not `httpx.HTTPStatusError` | Only the `httpx.HTTPStatusError` branch is caught — bare `AttributeError` propagates | **[CRASH]** FGA check produces 500, effectively fail-open by crashing before the fail-closed 403 path |

#### Summary
- Unhandled paths found: 4 (excluding non-issues)
- Critical (crash/data loss): 3 (shutdown-window None access in request handlers, revoke_share AttributeError, FGA dependency AttributeError)
- Non-critical (wrong result/degraded): 1 (missing-key error message only shows first missing var)

---

### Acceptance Auditor

#### Spec Extraction (from issue #216)

AC-1: Remove `init_cache_publisher`, `shutdown_cache_publisher`, `get_cache_publisher` globals from `cache_invalidation.py`
AC-2: Remove `_descope_client` global, `init_descope_client`, `shutdown_descope_client`, `get_descope_client` from `descope.py`
AC-3: Create instance in lifespan, store on `app.state.cache_publisher`; inject via `Depends(_get_publisher)`
AC-4: Create in lifespan, store on `app.state.descope_client`; inject via `Depends(_get_descope_client)`
AC-5: Remove `if self._publisher:` guards in all 6 services — null-object pattern guarantees publisher always present
AC-6: Fix operation labels: `sync` → `create` for new users, `sync` → `deactivate` for deletions, add `tenant_id` to role update publish
AC-7: Regenerate uv.lock (corrupt entry)

#### PASS

- [AC-1] `get_cache_publisher`, `init_cache_publisher`, `shutdown_cache_publisher` removed — implemented at `backend/app/services/cache_invalidation.py` (diff lines 567-614), tested via removal of `TestSingletonLifecycle` in `test_cache_invalidation.py`
- [AC-2] `_descope_client` global and all three functions removed — implemented at `backend/app/services/descope.py` (diff lines 630-663), tested via removal of `TestGetDescopeClient` in `test_descope_service.py`
- [AC-5] All 6 `if self._publisher:` guards removed — UserService, RoleService, PermissionService, TenantService, InboundSyncService, ReconciliationService. Null-object guarantee added at constructor level. Tested via updated service unit tests.
- [AC-6] Operation label fixes: `create`/`deactivate` in `inbound_sync.py` confirmed in diff; `tenant_id` added to role update publish confirmed. Tested implicitly through service method tests.

#### FAIL

- [AC-3] Spec requires "inject via `Depends(_get_publisher)`". The implementation does NOT use a `Depends()` helper function — it reads `request.app.state.cache_publisher` directly in each dependency factory (`get_user_service`, `get_role_service`, etc.). There is no `_get_publisher` function anywhere in the diff. The spec explicitly called for a `Depends()` indirection level. The implementation skips this abstraction, making the state coupling implicit and harder to override in tests without touching shared `app` state.
- [AC-4] Same issue as AC-3: spec requires "inject via `Depends(_get_descope_client)`". No such function exists. Direct `request.app.state.descope_client` attribute access is used instead. The `fga.py` dependency also accesses `request.app.state.descope_client` directly without any `Depends()` intermediary.

#### PARTIAL

- [AC-7] uv.lock regeneration cannot be verified from the diff alone — no `uv.lock` changes appear in the file list. PR description claims it was done. The prior review dismissed this as valid. Marking PARTIAL pending CI confirmation.

#### SCOPE CREEP

- `backend/tests/unit/conftest.py` — `_init_app_state` autouse fixture is an addition not mentioned in any AC. It is necessary for correctness (tests would break without it) but it introduces shared mutable state across tests.

#### Architecture Violations

- [D4 — Constructor injection] The architecture doc specifies constructor injection (`UserService(repository, adapter)` — FastAPI DI composes all layers at the edge). The example in the architecture doc still references `get_descope_client()` in the factory pattern. The PR correctly moves construction to the edge but the approach of bypassing `Depends()` intermediaries (AC-3/AC-4 failure) is inconsistent with the factory pattern documented in D4.

#### Summary
- Total ACs: 7
- Pass: 4 | Fail: 2 | Partial: 1

---

### Sentinel (Security)

#### BLOCK (must fix before merge)

No confirmed BLOCK-level vulnerabilities found from this refactor.

#### WARN (should fix)

- [LIKELY] `backend/app/main.py` (diff lines ~220-221, ~230) — Shutdown sets `app.state.descope_client = None` and `app.state.cache_publisher = None`. During the shutdown window, in-flight requests access these as `request.app.state.descope_client` directly in route handlers (`accesskeys.py`, `attributes.py`, `documents.py`, `fga.py`). An `AttributeError: 'NoneType'` will be raised. In the FGA dependency path (`dependencies/fga.py`), the check_permission call fails before reaching the fail-closed `raise HTTPException(403)` — the error propagates as a 500. This means FGA protection becomes effectively fail-open (crashes before access is denied) during the graceful shutdown window.
  Mitigation: Do not null `app.state` attributes on shutdown. Instead use a separate `_shutting_down` flag and fail requests with 503, or keep the NOOP instances in place.

- [UNLIKELY] `backend/app/dependencies/identity.py` — All dependency factory functions now accept `Request` as their first argument. FastAPI injects this automatically, but if any non-FastAPI caller invokes these functions directly (e.g., in a background task or script), they will receive `AttributeError` accessing `request.app.state`. Current usage appears to be FastAPI-only, so this is low risk.

#### INFO

- `backend/app/services/cache_invalidation.py` — Null-object pattern (`CacheInvalidationPublisher()` with `redis_client=None`) at construction time means cache invalidation silently fails without Redis. This is the intended degraded-mode behavior. Acceptable.
- `backend/tests/unit/conftest.py` — The autouse fixture sets `app.state.descope_client = AsyncMock()` on a shared `app` object. Individual tests overwrite this with specific mocks. There is no thread-safety concern in pytest's single-threaded async model, but test isolation is weaker than the previous `@patch` approach.

#### Summary
- BLOCK: 0 | WARN: 2 | INFO: 2
- Overall: PASS (with warnings that should be addressed before the shutdown-window race becomes a production incident)

---

### Viper (Red Team)

#### Attack Surface

- **All authenticated route handlers** (`accesskeys`, `attributes`, `documents`, `fga`, `fga dependencies`) — Changed from `get_descope_client()` (returns NOOP fallback or raises) to `request.app.state.descope_client` (returns `None` if lifespan hasn't run or has shut down). This opens a new crash path in the FGA authorization dependency.
- **FGA `require_fga` dependency** (`dependencies/fga.py`) — Changed from `get_descope_client()` to `request.app.state.descope_client`. FGA is the primary authorization enforcement point for document access.
- **Lifespan startup** — `DESCOPE_PROJECT_ID` / `DESCOPE_MANAGEMENT_KEY` now raise `RuntimeError` if missing, causing startup failure. Previously, `get_descope_client()` had a fallback that tried to read env vars on every call.

#### Findings

##### [MEDIUM] — FGA Authorization Fail-Open During Shutdown Window

- **Location**: `backend/app/dependencies/fga.py` (diff line ~19) + `backend/app/main.py` (diff line ~230)
- **Attack scenario**:
  1. Attacker observes a deployment/restart event (e.g., via HTTP 503 responses, CDN metrics, or social engineering)
  2. Attacker sends `GET /api/documents/{doc_id}` during the graceful-shutdown window (between SIGTERM and process exit)
  3. `request.app.state.descope_client` is `None` (set to `None` in shutdown path, diff line 230)
  4. `allowed = await client.check_permission(...)` becomes `None.check_permission(...)` → `AttributeError`
  5. `AttributeError` is not caught by `except httpx.HTTPStatusError` — propagates as 500
  6. FastAPI returns 500 before the fail-closed `raise HTTPException(status_code=400)` or permission-denied 403
  7. The document route handler that depends on FGA receives a 500 from the dependency, not a 403 — the request is rejected, so there is no direct access bypass. However, this is a reliability/availability attack vector: attacker can trigger 500s during shutdown by racing requests.
- **Prerequisites**: Timing knowledge of deployment window; no authentication required beyond a valid JWT (which attacker already has as a logged-in user)
- **CVSS v3.1**: 3.7 (AV:N/AC:H/PR:L/UI:N/S:U/C:N/I:N/A:L) — availability impact during shutdown only; no confidentiality or integrity impact because the 500 still blocks access
- **Remediation**: Do not null `app.state.descope_client` during shutdown. Keep the instance alive until the process exits (the HTTP client close is already the last step anyway).

##### [LOW] — Test Suite State Contamination Enables False Confidence

- **Location**: `backend/tests/unit/conftest.py` (diff lines ~1006-1022) + all test files that set `app.state.descope_client = mock_client`
- **Attack scenario**:
  1. Test A sets `app.state.descope_client = mock_client_A` (e.g., with `check_permission = True`)
  2. Test A crashes before the autouse fixture teardown runs
  3. Test B runs and the `_init_app_state` fixture resets to `AsyncMock()`, but Test B also sets `app.state.descope_client = mock_client_B`
  4. In a parallel test scenario (if ever enabled) or if test ordering changes, a mock from a previous test could leak into a security test, causing a test to pass that should fail (e.g., a permission-denied test that passes because it picks up a `check_permission = True` mock from a prior test)
  5. This is not an application security issue — but it creates false confidence in the test suite's security assertions
- **Prerequisites**: Test suite ordering change or pytest-xdist parallelism
- **CVSS v3.1**: 0.0 (not applicable — test-only) — flagged as process risk, not runtime vulnerability
- **Remediation**: Use the `@patch` approach or scope state per-test via `AsyncClient` with `app.state` set per-client, avoiding mutation of a shared module-level `app` object.

#### Summary
- Attack surface elements: 3
- Findings: 0 critical, 0 high, 1 medium, 1 low
- Overall: PASS (no exploitable auth bypass; medium finding is availability-only during deployment window)

---

### Triage summary

- **P0 (blocks merge)**: 2 — (1) Shutdown sets `app.state` attributes to `None` while in-flight requests access them directly, causing `AttributeError` crashes in route handlers including the FGA `check_permission` dependency (fail-open crash path); (2) AC-3 and AC-4 from spec are not implemented — spec required `Depends(_get_publisher)` and `Depends(_get_descope_client)` intermediaries; implementation reads `request.app.state.*` directly in factories, skipping the `Depends()` abstraction contract
- **P1 (should fix)**: 3 — (1) `InboundSyncService`/service constructor `publisher` type annotations still say `| None` when nulls are no longer possible; (2) `revoke_share` `try/except` wraps `request.app.state.descope_client` access with a stale "Failed to initialize client" log message — cleanup needed; (3) autouse `_init_app_state` fixture introduces shared mutable state; stronger isolation recommended
- **P2 (nitpick)**: 1 — Missing-env-var `RuntimeError` message says both vars but only surfaces which one triggered the KeyError

The core refactor intent (remove module-level globals, wire via lifespan) is sound and correctly implemented. The P0 on spec AC-3/AC-4 is a judgment call — the `Depends()` intermediaries were listed in the spec but the direct `request.app.state.*` access is functionally equivalent and arguably simpler. However the shutdown-window None-setting creating crash paths in FGA handlers is a genuine reliability bug that should be fixed before this lands. Recommend: (1) remove the `app.state.X = None` lines from shutdown (let the objects live until process exit), (2) decide whether the `Depends()` abstraction from the spec is required or can be formally waived as PARTIAL acceptance.

Issue #165, opened 2026-03-31
Epic 1 Story 1.5: Configure Health Check Endpoint Passthrough


jamescrowley321:
## Story 1.5: Configure Health Check Endpoint Passthrough

As a developer operating the gateway,
I want the `/api/health` endpoint to be accessible without authentication through Tyk,
So that health checks and readiness probes work without a JWT.

**Acceptance Criteria:**

- [ ] The Tyk API definition includes a whitelist or `ignored_paths` entry for `/api/health` that bypasses JWT validation
- [ ] `GET http://localhost:8080/api/health` returns 200 without an `Authorization` header when the gateway profile is running
- [ ] All other `/api/*` endpoints still require a valid JWT
- [ ] The standalone mode health check behavior is unchanged (`GET http://localhost:8000/api/health` returns 200)

**Depends on:** Story 1.4

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 2 (API Gateway) shipped in full on 2026-04-11/12 via PRs #210-#239.

PR #229, opened 2026-04-10
feat: health check passthrough in Tyk gateway (#165)

jamescrowley321:
## Summary

- Added `ignored` entry in Tyk `extended_paths` for `/api/health` GET — bypasses JWT validation so health/readiness probes work without authentication through the gateway
- Updated integration tests: `test_health_proxy` expects 200 always (removed 401-skip fallback), JWT rejection tests moved to `/api/me`
- Added `test_health_no_auth_via_gateway` (sends garbage auth token to prove auth is truly *ignored*, not just optional) and `test_tyk_config_health_ignored` (config verification)
- Shell test script updated: AC1 expects 200 always, AC4 tests invalid JWT against `/api/me`

Refs #165

## Review Findings Addressed

5 reviewers spawned (Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel, Viper).

- **1 CRITICAL, 2 HIGH** (Blind Hunter): dismissed with justification — unauthenticated health is the story's purpose, Sentinel confirmed safe, FastAPI enforces GET-only via 405
- **1 HIGH fixed** (H2): `test_health_no_auth_via_gateway` strengthened to send garbage auth token proving auth is truly ignored
- **3 MEDIUM** (Blind Hunter) + **1 MEDIUM** (Viper): dismissed — `/api/me` exists, pre-flight logic works, HEAD not in scope, FastAPI enforces method restrictions
- **3 LOW/INFO**: skipped per P2 triage rules
- **Acceptance Auditor**: 4/4 AC PASS, no scope creep
- **Sentinel**: PASS, 0 BLOCK, 0 WARN
- **Edge Case Hunter**: 0 findings

## Test plan
- [x] Unit tests pass (1584)
- [x] Lint passes
- [x] JSON validated
- [ ] CI passes


jamescrowley321:
# Blind Hunter Review: review-diff.patch

Reviewer: The Blind Hunter
Date: 2026-04-10
Verdict: **Several concerns, one security-critical**

---

## CRITICAL

### C1: Health endpoint exempt from authentication creates an unauthenticated oracle

**Files:** `tyk/apps/saas-backend.json`, `scripts/test-gateway-proxy.sh`, `backend/tests/integration/test_gateway_proxy.py`

The entire point of this diff is to make `/api/health` bypass JWT validation at the gateway layer. The Tyk config adds it as an `"ignored"` path with `"action": "no_action"`.

This means `/api/health` is now fully unauthenticated and publicly reachable through the gateway. Depending on what the health endpoint returns, this can leak:
- Backend version/build info
- Database connectivity status (up/down signals useful for timing attacks)
- Internal service names or dependency health
- Whether the system is degraded (useful for coordinated attacks)

The diff does not show what `/api/health` actually returns, but the test asserts `data["status"] == "ok"`, suggesting it at minimum confirms backend liveness. If it returns anything beyond a static `{"status": "ok"}`, this is an information disclosure risk.

**Additionally:** The `"ignored"` path only specifies `"GET"` in `method_actions`, but Tyk's `ignored` path behavior may vary by version. Some Tyk versions ignore the `method_actions` filter entirely on `ignored` paths, meaning POST/PUT/DELETE to `/api/health` would also bypass auth. This needs verification against the deployed Tyk version.

**Recommendation:** Confirm the health endpoint returns only a static response. Verify Tyk actually enforces the GET-only restriction on ignored paths. Consider rate-limiting this endpoint even though it is auth-exempt.

---

## HIGH

### H1: Removing the 401-tolerant fallback makes the test suite and shell script fragile

**Files:** `scripts/test-gateway-proxy.sh` (lines 107-113), `backend/tests/integration/test_gateway_proxy.py` (lines 9-19)

The previous code gracefully handled the case where `/api/health` required auth (returned 401) by skipping the test or noting the info. The new code hard-asserts 200. If the Tyk config change fails to deploy, is overridden by a higher-priority policy, or the `ignored` path doesn't work as expected in a particular Tyk version, the entire test suite fails with a cryptic assertion error instead of a helpful diagnostic.

The old behavior was resilient. The new behavior is brittle. This is a regression in test robustness.

### H2: `test_health_proxy` and `test_health_no_auth_via_gateway` are duplicates

**File:** `backend/tests/integration/test_gateway_proxy.py`

`test_health_proxy` (line 8) and `test_health_no_auth_via_gateway` (line 24) do the exact same thing: issue an unauthenticated GET to `/api/health` and assert 200. The first one also checks `data["status"] == "ok"`, and the second does not. They share the same logical assertion. This is dead test code masquerading as a separate acceptance criterion.

If these are meant to test different things, the second test adds nothing. If the intent is "AC1 = proxies correctly" and "AC2 = no auth needed," they should be a single test, or the second should explicitly verify that adding an Authorization header also works (to prove auth is truly *ignored*, not just *optional*).

---

## MEDIUM

### M1: JWT rejection tests now depend on `/api/me` existing and being protected

**Files:** `backend/tests/integration/test_gateway_proxy.py` (lines 37-38, 46-47), `scripts/test-gateway-proxy.sh` (lines 121-127)

The invalid/expired JWT tests were changed from hitting `/api/health` to `/api/me`. This is correct in principle (you can't test auth rejection on an auth-exempt path), but now these tests have an implicit dependency on `/api/me` existing as a protected route in the Tyk config.

If `/api/me` doesn't exist in the backend, the test might get a 404 instead of 401/403, and the assertion `response.status_code in (401, 403)` would fail with a misleading error. There is no guard or skip logic for this case. The diff does not show `/api/me` being added to the Tyk config or backend routes, so this is either an existing route (fine) or a missing dependency (broken tests).

### M2: Shell script pre-flight logic has a dead fallback path

**File:** `scripts/test-gateway-proxy.sh` (lines 79-87)

The pre-flight check uses `curl -sf` which fails on non-2xx responses. Since `/api/health` is now auth-exempt and should return 200, the `else` branch (HTTP_CODE fallback check) should never trigger in normal operation. The fallback that checks for non-000 codes and says "Gateway responded with HTTP $HTTP_CODE (may require auth)" is now misleading -- the health endpoint should never require auth. This comment/logic is stale and contradicts the new design.

### M3: Only GET is ignored -- HEAD requests to `/api/health` will be auth-gated

**File:** `tyk/apps/saas-backend.json`

The `method_actions` only lists `"GET"`. Monitoring systems and load balancers commonly use HEAD requests for health checks. If any infrastructure component sends a HEAD to `/api/health`, it will hit the JWT validation and fail. This may or may not matter, but it's worth noting.

---

## LOW

### L1: Trap overwrite risk in shell script

**File:** `scripts/test-gateway-proxy.sh` (line 96)

```
trap 'rm -f "$HEALTH_TMPFILE"' EXIT
```

This sets a new EXIT trap. If there was a prior EXIT trap set earlier in the script (or set later), this one overwrites it (or gets overwritten). In bash, `trap` replaces the previous handler for the same signal. If other temp files are created later with their own `trap ... EXIT`, this cleanup could be silently dropped. The script should accumulate cleanup in a single trap or use a cleanup function.

### L2: Test `test_tyk_config_health_ignored` does deep dict traversal without defensive checks

**File:** `backend/tests/integration/test_gateway_proxy.py` (lines 55-65)

The chain `.get("version_data", {}).get("versions", {}).get("Default", {}).get("extended_paths", {})` silently returns `{}` at any broken link, which then returns `ignored = []`, which then fails the assertion with "Expected exactly one ignored entry for /api/health." This is actually fine for a test -- the assertion message is clear enough -- but the test doesn't tell you *where* the config is broken in the chain. A missing `version_data` key gives the same error as a missing `ignored` entry.

### L3: Hardcoded timeout of 10.0 seconds across all async tests

**File:** `backend/tests/integration/test_gateway_proxy.py`

Every `httpx.AsyncClient` uses `timeout=10.0`. This is fine for local testing but may be too short for CI environments with cold starts, or too long for fast-fail scenarios. This was pre-existing, not introduced by this diff, but the new tests copy the same pattern.

---

## Summary

| Severity | Count | Key Issue |
|----------|-------|-----------|
| CRITICAL | 1 | Unauthenticated health endpoint may leak info; GET-only restriction on ignored paths needs Tyk version verification |
| HIGH | 2 | Removed resilient fallback logic; duplicate test |
| MEDIUM | 3 | Implicit `/api/me` dependency; stale shell logic; HEAD method not ignored |
| LOW | 3 | Trap overwrite; opaque config traversal failure; hardcoded timeouts |

**Total findings: 9**

The diff is small and focused, but it makes a security-relevant change (bypassing auth on a path) with insufficient verification that the bypass is properly scoped. The test changes are logically correct but introduce fragility and duplication.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

No unhandled edge cases found.

All branching paths in the changed code are properly guarded:

- **`saas-backend.json`**: The `ignored` path entry is structurally valid. The `method_actions` specifies `GET` with `no_action`. Other HTTP methods hitting `/api/health` bypass Tyk auth but receive 405 from FastAPI (only GET is registered), so no auth bypass is exploitable.
- **`test_gateway_proxy.py`**: The `test_tyk_config_health_ignored` test uses safe `.get()` chains with `{}` defaults throughout. `health_paths[0]` access is guarded by the preceding `len == 1` assertion. `response.json()` in `test_health_proxy` is preceded by a `status_code == 200` assertion, and a `JSONDecodeError` would correctly surface as a test failure.
- **`test-gateway-proxy.sh`**: The removal of the `401` skip path in AC1 is intentional — if the ignored path config is not working, the test should hard-fail. The pre-flight section still handles unreachable gateways (`HTTP_CODE == 000`) and non-200 codes (proceeds with a warning, same as before).

### Summary
- Unhandled paths found: 0

jamescrowley321:
## Review: Acceptance Auditor

**Spec:** GitHub Issue #165 — Epic 1 Story 1.5: Configure Health Check Endpoint Passthrough
**Diff:** `791d220..0c2d059` (1 commit: `feat: add health check passthrough in Tyk gateway config`)

- **AC-1:** The Tyk API definition includes a whitelist or `ignored_paths` entry for `/api/health` that bypasses JWT validation
- **AC-2:** `GET http://localhost:8080/api/health` returns 200 without an `Authorization` header when the gateway profile is running
- **AC-3:** All other `/api/*` endpoints still require a valid JWT
- **AC-4:** The standalone mode health check behavior is unchanged (`GET http://localhost:8000/api/health` returns 200)

---

### PASS

- **[AC-1]** The Tyk API definition includes an `ignored` entry for `/api/health` that bypasses JWT validation — implemented at `tyk/apps/saas-backend.json:49-58` (added `"ignored"` array with `/api/health` path, `GET` method, `"action": "no_action"`). Tested by:
  - Config unit test: `backend/tests/integration/test_gateway_proxy.py:110-120` (`test_tyk_config_health_ignored`) — verifies the JSON structure has exactly one ignored entry for `/api/health` with the correct `method_actions`.
  - Integration test: `backend/tests/integration/test_gateway_proxy.py:41-49` (`test_health_proxy`) — verifies `GET /api/health` returns 200 through the gateway (no longer accepts 401 with `pytest.skip`, now asserts 200 strictly).
  - Shell script: `scripts/test-gateway-proxy.sh:41-58` — AC1 section now asserts 200, removed the old fallback that accepted 401.

- **[AC-2]** `GET http://localhost:8080/api/health` returns 200 without an `Authorization` header when the gateway profile is running — implemented by the `ignored` path config in `tyk/apps/saas-backend.json:49-58`. Tested by:
  - Integration test: `backend/tests/integration/test_gateway_proxy.py:52-57` (`test_health_no_auth_via_gateway`) — explicitly sends no `Authorization` header and asserts 200.
  - Integration test: `backend/tests/integration/test_gateway_proxy.py:41-49` (`test_health_proxy`) — also sends no `Authorization` header and asserts 200.
  - Shell script: `scripts/test-gateway-proxy.sh:46-58` — curl with no auth headers, asserts 200.

- **[AC-3]** All other `/api/*` endpoints still require a valid JWT — the `ignored` list in `tyk/apps/saas-backend.json` only contains `/api/health`; all other `/api/*` paths are still covered by the `use_openid: true` authentication policy. Tested by:
  - Integration tests that previously tested JWT rejection against `/api/health` were correctly updated to test against `/api/me` (a protected endpoint):
    - `backend/tests/integration/test_gateway_proxy.py:60-68` (`test_invalid_jwt_rejected`) — sends invalid JWT to `/api/me`, asserts 401/403.
    - `backend/tests/integration/test_gateway_proxy.py:71-80` (`test_expired_jwt_rejected`) — sends expired JWT to `/api/me`, asserts 401/403.
    - `backend/tests/integration/test_gateway_proxy.py:83-87` (`test_missing_auth_rejected`) — no auth to `/api/me`, asserts 401/403.
    - `backend/tests/integration/test_gateway_proxy.py:90-97` (`test_empty_auth_header_rejected`) — empty auth to `/api/me`, asserts 401/403.
    - `backend/tests/integration/test_gateway_proxy.py:100-107` (`test_malformed_bearer_rejected`) — malformed bearer to `/api/me`, asserts 401/403.
  - Shell script: `scripts/test-gateway-proxy.sh:97-108` — AC4 invalid JWT test now targets `/api/me` instead of `/api/health`.
  - Shell script: `scripts/test-gateway-proxy.sh:111-122` — AC5 missing auth test targets `/api/me`.

- **[AC-4]** The standalone mode health check behavior is unchanged — **zero backend application code was modified** in this diff (confirmed via `git diff 791d220..0c2d059 -- backend/app/` producing empty output). The health endpoint at `backend/app/routers/health.py` is untouched. Pre-existing tests confirm standalone behavior:
  - Unit test: `backend/tests/unit/test_health.py:22-25` (`test_health_returns_ok`) — asserts 200 with `{"status": "ok"}`.
  - Unit test: `backend/tests/unit/test_health.py:29-32` (`test_health_no_auth_required`) — asserts 200 without auth.
  - E2E test: `backend/tests/e2e/test_api_health.py:6-11` (`test_health_endpoint_returns_ok`) — standalone endpoint returns 200.
  - E2E test: `backend/tests/e2e/test_api_health.py:14-17` (`test_health_endpoint_no_auth_required`) — no auth required in standalone.
  - The middleware factory at `backend/app/middleware/factory.py:67-68` already had `/api/health` in `excluded_paths` for `TokenValidationMiddleware` in standalone mode — unchanged by this diff.

### FAIL

(none)

### PARTIAL

(none)

### SCOPE CREEP

(none detected)

All changes in the diff are directly traceable to the four acceptance criteria:

1. `tyk/apps/saas-backend.json:49-58` — AC-1 implementation (ignored path config).
2. `backend/tests/integration/test_gateway_proxy.py` — test updates for AC-1 through AC-3 (new `test_health_no_auth_via_gateway`, new `test_tyk_config_health_ignored`, fixed JWT rejection tests to target `/api/me` instead of `/api/health`).
3. `scripts/test-gateway-proxy.sh` — shell script updates aligning with the auth-exempt behavior (removed 401 fallback acceptance for `/api/health`, retargeted AC4/AC5 JWT tests to `/api/me`).

The retargeting of JWT rejection tests from `/api/health` to `/api/me` is not scope creep — it is a necessary correction because `/api/health` is now auth-exempt, so testing JWT rejection against it would no longer be valid.

### Architecture Violations

(none)

- **ADR-GW-3 (File-Based API Definitions):** Compliant. The change is made in the file-based Tyk API definition `tyk/apps/saas-backend.json`, which is version-controlled.
- **Architecture Section 6 (Tyk Configuration Architecture):** The `ignored` paths mechanism in `extended_paths` is the standard Tyk approach for bypassing authentication on specific paths. The implementation uses `"use_extended_paths": true` which was already present.
- **Architecture Section 8 (Security Architecture):** The health endpoint bypass does not weaken the security posture. Only the read-only, non-sensitive `/api/health` endpoint (GET only) is exempted. All other endpoints remain protected by OpenID Connect JWT validation.

### Summary

- Total ACs: 4

jamescrowley321:

No blocking findings.

### WARN (should fix)

No warnings.

### INFO (acceptable risk)

- `tyk/apps/saas-backend.json:49-58` — `/api/health` added as a Tyk "ignored" path, bypassing JWT validation at the gateway. The endpoint returns only `{"status": "ok"}` with no sensitive data, no user input processing, and no database access. The backend route is GET-only (FastAPI returns 405 for other methods). In standalone mode, `/api/health` was already excluded from `TokenValidationMiddleware` via `excluded_paths`. This is consistent and acceptable.

- `backend/tests/integration/test_gateway_proxy.py:37,47` — JWT rejection tests (AC4/AC5) now target `/api/me` instead of `/api/health`. This is the correct change since `/api/health` is now auth-exempt, so testing JWT rejection against it would be meaningless. `/api/me` depends on `get_current_user` which checks `request.state.principal`, providing a proper protected endpoint for these tests.

- `tyk/apps/saas-backend.json:49-58` — The Tyk "ignored" path configuration specifies `method_actions` for GET only. In Tyk's implementation, when a path is in the `ignored` list, authentication is bypassed for that path regardless of the `method_actions` content (the `method_actions` field controls response behavior, not auth gating). However, the backend only registers a GET handler at `/health`, so non-GET requests receive a 405 from FastAPI. No exploitable gap exists.

### Summary
- Overall: PASS

jamescrowley321:
## Review: Red Team (Viper)

### Attack Surface
- **`tyk/apps/saas-backend.json` — "ignored" path added for `/api/health`**: The Tyk API definition now includes `/api/health` in the `extended_paths.ignored` list, which causes Tyk to skip JWT validation for requests to that path. This is the core security-relevant change.
- **Test files updated**: Integration tests and the shell script now assert `/api/health` returns 200 without auth, and JWT rejection tests moved from `/api/health` to `/api/me`. These are test-only changes and do not alter the attack surface directly, but they confirm the intended behavior.

### Findings

#### MEDIUM — Tyk "ignored" path bypasses auth for ALL HTTP methods, not just GET

- **Location**: `tyk/apps/saas-backend.json:49-58`
- **Attack scenario**:
  1. Attacker sends `POST /api/health`, `PUT /api/health`, `DELETE /api/health`, or `OPTIONS /api/health` through the Tyk gateway without any JWT.
  2. Tyk's "ignored" path mechanism bypasses all authentication middleware for the path `/api/health` regardless of HTTP method. While `method_actions` specifies `GET` with `no_action`, Tyk's ignored paths are path-level exemptions — the `method_actions` field controls what Tyk *does* for each method (reply, redirect, no_action), not whether auth is enforced. All methods on the ignored path skip authentication.
  3. The backend's FastAPI health endpoint only registers `GET /api/health`, so non-GET methods will return 405 Method Not Allowed from FastAPI. This means the actual exploitability is limited — there is no code execution path for POST/PUT/DELETE on `/api/health`.
  4. However, if a future developer adds a POST or PUT handler on `/api/health` (e.g., a health check that accepts configuration, or a readiness toggle), it would be unauthenticated through the gateway with zero additional Tyk config changes.
- **Prerequisites**: Network access to the Tyk gateway (port 8080)
- **CVSS v3.1**: 3.7 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:L/A:N) — currently unexploitable (no non-GET handler exists), but the config creates latent risk
- **Remediation**: This is defense-in-depth. The current config is acceptable because FastAPI only exposes GET on `/api/health`. To harden: no code change needed now, but document in the Tyk config that any new method on `/api/health` will bypass gateway auth. Consider adding a comment in `saas-backend.json` noting this behavior.

#### LOW — Health endpoint information disclosure through gateway (intentional but worth noting)

- **Location**: `tyk/apps/saas-backend.json:49-58`, `backend/app/routers/health.py:8-11`
- **Attack scenario**:
  1. Attacker sends `GET /api/health` to the Tyk gateway without authentication.
  2. Receives `{"status": "ok"}` confirming the backend is running and reachable.
  3. This confirms: (a) Tyk gateway is operational, (b) the backend service is running, (c) the backend is reachable from Tyk's network.
  4. An attacker can use this for reconnaissance — continuously polling to detect deployment windows, downtime patterns, or confirm the target is alive before launching further attacks.
- **Prerequisites**: Network access to the Tyk gateway
- **CVSS v3.1**: 0.0 — This is intentional behavior (health checks must be unauthenticated for monitoring). Not a vulnerability per se, but documenting for completeness.
- **Remediation**: None required. This is standard practice. The health endpoint returns minimal information (no version, no internal state, no dependency status). If richer health data is added in the future (e.g., database status, cache status, version info), consider splitting into a public liveness probe and a protected readiness probe.

### Non-Findings (Investigated and Cleared)

**Backend defense-in-depth is intact**: In standalone mode, `TokenValidationMiddleware` already excludes `/api/health` from JWT validation (line 68 of `factory.py`). In gateway mode, `TokenValidationMiddleware` is entirely skipped — Tyk is the auth boundary. The Tyk ignored path change aligns the gateway-mode behavior with what standalone mode already does. No new unauthenticated surface is created that did not already exist.

**No auth bypass on protected endpoints**: The tests correctly moved JWT rejection assertions from `/api/health` (now auth-exempt) to `/api/me` (protected). `/api/me` depends on `get_current_user` which checks `request.state.principal` — in gateway mode this is currently fail-closed (returns 401) since the claim-extraction middleware is not yet implemented. In standalone mode, `TokenValidationMiddleware` validates the JWT.

**No method confusion on the backend**: FastAPI's router only registers GET for `/api/health`. Non-GET methods return 405 automatically. There is no path traversal or path normalization concern — Tyk matches the exact path `/api/health`.

**No infrastructure escape via the ignored path**: The health endpoint returns a static JSON object with no user-controlled input reflected in the response. No SSRF, no injection, no data leak.

**Internal endpoints remain protected**: The `/api/internal/` prefix is excluded from JWT auth in standalone mode but protected by per-endpoint shared secrets (HMAC, X-Flow-Secret, X-Identity-Key). In gateway mode, Tyk does NOT have an ignored path for `/api/internal/` — Tyk will require a JWT for those paths. The internal endpoints have their own secret-based auth regardless.

**Docker network topology is unchanged**: Backend is bound to `127.0.0.1:8000` on the host, Tyk gateway to `127.0.0.1:8080`. Within the Docker network, backend is reachable at `backend:8000` from any container, but this is the pre-existing topology — this PR does not change it.

### Summary
- Attack surface elements: 1 (Tyk ignored path for `/api/health`)
- Findings: 0 critical, 0 high, 1 medium, 1 low
- Overall: **PASS** — The change is a correct and expected configuration for exempting a health check endpoint from JWT validation at the gateway layer. The medium finding is a defense-in-depth observation about Tyk's method-agnostic behavior on ignored paths, not an actively exploitable vulnerability. The backend already treated `/api/health` as unauthenticated in standalone mode, so this change brings gateway mode into alignment.

jamescrowley321:

This is a re-review conducted under fresh context per the `feedback_ralph_reviews_shallow` guidance. Five reviewer personas applied against the full diff, the spec (issue #165), the architecture doc, and the codebase on disk.

---

### Blind Hunter

Findings from the diff alone, no project context assumed.

**MUST FIX**

No findings.

**SHOULD FIX**

- [`tyk/apps/saas-backend.json:146-155`] The `ignored` entry uses `"action": "no_action"` with no `code` or `headers` sub-fields. In Tyk OSS, `method_actions` entries with `"action": "no_action"` are correctly interpreted as pass-through — but this is an unusual pattern and the absence of the complementary `reply_action` and `data` fields could silently no-op on older Tyk versions (pre-4.x) that require `"action": "reply"` for proper passthrough in extended paths. The diff provides no documentation of the minimum Tyk version required for this config shape.

- [`backend/tests/integration/test_gateway_proxy.py:25`] The `gateway_url` fixture (unchanged in this PR, but materially affected by it) probes reachability via `httpx.get(f"{TYK_URL}/api/health")` and accepts any response — including `401` — as "gateway is reachable." After this PR, `/api/health` always returns `200`. However, the fixture comment still says "even a 401 means it's up," which is now stale and misleading for the next reader.

- [`backend/tests/integration/test_gateway_proxy.py:62-73`] `test_tyk_config_health_ignored` verifies the static JSON config file on disk, not the running gateway. If the Tyk process loaded a stale config (e.g., container not restarted after editing the JSON), this test passes but the live gateway still enforces auth on `/api/health`. The test is useful but its scope limitation is not commented.

**NITPICK**

- [`scripts/test-gateway-proxy.sh:86-91`] The pre-flight block still has a dead code path: after the `ignored` entry is deployed, `curl -sf` on `/api/health` will always succeed (exit 0), so the inner `if [ "$HTTP_CODE" = "000" ]` branch is now unreachable under normal operation. Not harmful, just permanently skipped code in the happy path.

---

### Edge Case Hunter

Analysis of branching paths and boundary conditions.

| `test_gateway_proxy.py:24-36` (new `test_health_no_auth_via_gateway`) | Gateway is up but running with stale config (no `ignored` entry loaded yet — container not restarted) | Add comment: `# Requires gateway restart after config change` | **[WRONG]** Test fails with 401; developer sees false failure, may incorrectly conclude the ignored-path config is broken |
| `tyk/apps/saas-backend.json:147` | `method_actions` key is `"GET"` only; a `HEAD /api/health` request is not listed | Add `"HEAD": {"action": "no_action"}` if needed | **[DEGRADED]** Kubernetes/AWS ALB liveness probes that use `HEAD` instead of `GET` bypass the ignored-path entry and hit JWT validation, returning 401 — health check fails |
| `test_gateway_proxy.py:62-73` (`test_tyk_config_health_ignored`) | `saas-backend.json` is loaded but `envsubst` was not run (template vars like `${DESCOPE_PROJECT_ID}` still present) | n/a — file parses as valid JSON regardless | **[WRONG]** `json.load(f)` succeeds; `extended_paths.ignored` is present; test passes — but the live Tyk gateway rejects the config entirely, meaning the ignored path is never active |

### Summary
- Unhandled paths found: 3

---

### Acceptance Auditor

Spec: issue #165 — Story 1.5: Configure Health Check Endpoint Passthrough

ACs extracted:
- **AC-1**: Tyk API definition includes a whitelist or `ignored_paths` entry for `/api/health` that bypasses JWT validation
- **AC-2**: `GET http://localhost:8080/api/health` returns 200 without an `Authorization` header when gateway profile is running
- **AC-3**: All other `/api/*` endpoints still require a valid JWT
- **AC-4**: Standalone mode health check behavior is unchanged (`GET http://localhost:8000/api/health` returns 200)

**PASS**

- [AC-1] `ignored` entry added for `/api/health` GET in `extended_paths` — implemented at `tyk/apps/saas-backend.json:146-155`, config-verified at `test_gateway_proxy.py:62-73` (`test_tyk_config_health_ignored`)

- [AC-2] Integration test enforces 200 without auth — implemented at `test_gateway_proxy.py:11-21` (updated `test_health_proxy` removes 401-skip fallback) and `test_gateway_proxy.py:24-36` (`test_health_no_auth_via_gateway`)

- [AC-3] JWT rejection tests now target `/api/me` (a genuinely protected endpoint) — implemented at `test_gateway_proxy.py:39-57` and `scripts/test-gateway-proxy.sh:128-137`. The shift from `/api/health` to `/api/me` for AC4 JWT rejection is correct and stronger.

- [AC-4] No changes to `backend/app/middleware/factory.py`, `backend/app/routers/health.py`, or `backend/app/middleware/auth.py`. Standalone mode `excluded_paths` already contains `/api/health` at `factory.py:68`. Standalone health behavior is untouched.

**FAIL**

No failures.

**PARTIAL**

No partial implementations.

**SCOPE CREEP**

None detected. All diff changes are traceable to issue #165.

**Architecture Violations**

None. The `ignored` path entry is a Tyk-layer concern (authentication boundary). FastAPI still owns authorization. ADR-GW-1 boundary is intact.

### Summary
- Total ACs: 4

---

### Sentinel (Security)

**BLOCK (must fix before merge)**

No BLOCK findings.

**WARN (should fix)**

- [LIKELY] `tyk/apps/saas-backend.json:146-155` — The `ignored` entry bypasses JWT validation for `GET /api/health`. Tyk's `ignored` path behavior also bypasses rate limiting for that path in addition to auth. This is consistent with intent, but it means `/api/health` is now fully exempt from the global 60 req/min rate limit as well. A high-volume automated scanner or misconfigured probe could use this path to generate load on the backend without triggering Tyk's rate limiter.
  Mitigation: Acceptable for a health endpoint serving k8s-style probes (low volume, fixed-source). Worth noting explicitly in the API definition's `x-description` field. Not a block.

- [UNLIKELY] `backend/app/middleware/factory.py:63-77` — In gateway mode, `TokenValidationMiddleware` is excluded from the stack. The `ignored` entry in Tyk means `/api/health` bypasses Tyk JWT validation. The FastAPI health route itself is already `@limiter.exempt` and has no auth dependency. This is correct — but if a future change adds auth to the health handler (e.g., adding `Depends(get_current_user)`), the gateway-mode path would silently fail because there is no token on the request. Not exploitable today; risk only if the health handler is inadvertently made auth-aware in the future.
  Mitigation: Add a comment on the health route explaining it must remain auth-free in all deployment modes.

**INFO**

- `backend/app/middleware/factory.py:68` — `/api/health` is already in `excluded_paths` for `TokenValidationMiddleware` in standalone mode. This is consistent with the Tyk `ignored` entry: both modes treat the health endpoint as auth-exempt. Defense-in-depth is correctly maintained.

### Summary
- BLOCK: 0 | WARN: 2 | INFO: 1
- Overall: PASS

---

### Viper (Red Team)

Activated: changes touch Tyk gateway configuration (infrastructure/auth boundary), middleware passthrough routing, and integration test auth enforcement.

**Attack Surface**

- [`tyk/apps/saas-backend.json`] — New `ignored` path entry creates an unauthenticated passthrough route through the JWT validation layer
- [`test_gateway_proxy.py`] — JWT rejection tests moved from `/api/health` to `/api/me`; changes what the test suite actually exercises against the live gateway

**Stage 1: Recon**

Three attack surface elements:
1. New unauthenticated path: `GET /api/health` — no JWT required, passes through Tyk entirely
2. The `ignored` path only applies to the Tyk layer; FastAPI's health handler has no auth dependency (`@limiter.exempt` only)
3. Rate limiting is also bypassed on the ignored path (Tyk-level rate limit does not apply to ignored paths)

**Stage 2 & 3: Vulnerability Analysis and Findings**

#### [LOW] — Health path bypasses gateway rate limiting (not just auth)

- **Location**: `tyk/apps/saas-backend.json:146-155`
- **Attack scenario**:
  1. Attacker identifies `GET /api/health` is exempt from Tyk auth and rate limiting
  2. Attacker sends 10,000 requests/minute to `GET /api/health` — Tyk forwards all of them to FastAPI without rate-limit enforcement
  3. FastAPI's `@limiter.exempt` decorator means slowapi also does not limit this endpoint
  4. Backend sustains sustained load from a single unauthenticated source
- **Prerequisites**: Network access to the gateway port (8080). No credentials required.
- **CVSS v3.1**: 3.7 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L) — Low, because: (a) it requires sustained effort to cause meaningful impact on a lightweight endpoint, (b) the health handler returns a static dict with zero database access, (c) any real DDoS protection would be at the network/infra layer, not the gateway
- **Remediation**: Add a separate rate-limit rule for `/api/health` in `extended_paths.rate_limit` if flood protection is desired. For a reference architecture this is acceptable risk.

#### [MEDIUM] — JWT rejection tests now cover a narrower blast radius than before

- **Location**: `backend/tests/integration/test_gateway_proxy.py:39-57`
- **Attack scenario**: This is a test coverage gap rather than a direct exploit. The previous tests sent invalid/expired JWTs to `/api/health` — which, though wrong in intent (health should be exempt), exercised the Tyk JWT rejection path on the same listen path that all API traffic uses. The new tests send to `/api/me`. If Tyk's `ignored` path processing has a bug where it applies the ignore rule too broadly (e.g., path-prefix matching rather than exact matching), a request to `/api/health-admin` would also bypass auth, and the current test suite would not catch it.
  - Step 1: Attacker notes Tyk is running OSS v5.x with `ignored` path entry for `/api/health`
  - Step 2: Attacker sends `GET /api/healthz` or `GET /api/health/details` (plausible future paths) without auth
  - Step 3: If Tyk uses prefix matching on the `ignored` entry, request passes through without JWT validation
  - Step 4: FastAPI in gateway mode has no `TokenValidationMiddleware`, so the request reaches the route handler unauthenticated
- **Prerequisites**: Knowledge that an `ignored` path entry exists; access to gateway port; a future endpoint under `/api/health*` prefix
- **CVSS v3.1**: 5.3 (AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:L/A:N) — Medium, contingent on Tyk doing prefix matching (which it does NOT do for `ignored` paths; Tyk uses exact path matching by default). Scoring accounts for uncertainty about Tyk's exact behavior.
- **Remediation**: Add an integration test that verifies a path like `GET /api/healthz` (no ignored entry) still returns 401 when sent without auth. This would catch any prefix-match regression in Tyk config.

**Note on the MEDIUM finding**: Tyk OSS documentation specifies exact path matching for `ignored` entries, not prefix matching. This finding is flagged as a test coverage gap rather than a confirmed vulnerability. If the team has verified Tyk's exact-match behavior, this can be downgraded to LOW/INFO.

### Summary
- Attack surface elements: 3
- Findings: 0 critical, 0 high, 1 medium, 1 low
- Overall: PASS

---

### Triage summary

- **P0 (blocks merge)**: 0 — none
- **P1 (should fix before or shortly after merge)**: 3
  1. `test_gateway_proxy.py:25` — fixture comment "even a 401 means it's up" is now stale/misleading (Blind Hunter SHOULD FIX)
  2. `tyk/apps/saas-backend.json:147` — `HEAD /api/health` not in `method_actions`; breaks `HEAD`-based liveness probes (Edge Case Hunter, DEGRADED)
  3. Viper MEDIUM — no test verifying that `/api/healthz` or similar paths are NOT on the ignored list (test coverage gap against future prefix-match regressions)
- **P2 (nitpick / low)**: 3
  - Shell pre-flight dead code path (Blind Hunter)
  - Sentinel WARN: rate limiting also bypassed on ignored path (acceptable risk, worth documenting)
  - Edge Case Hunter: config-file test passes even when `envsubst` not run

All four ACs pass. No security blocks. The two P1 items (stale fixture comment and missing `HEAD` method in `method_actions`) are low-risk for this story but worth tracking. The `HEAD` issue is the most operationally significant: if any health-check probe or ALB target-group health check uses `HEAD` rather than `GET`, the ignored-path entry will not apply and the probe will receive a 401.

Issue #169, opened 2026-03-31
Epic 2 Story 2.4: Verify Authorization Independence from Deployment Mode


jamescrowley321:
## Story 2.4: Verify Authorization Independence from Deployment Mode

As a developer ensuring authorization correctness,
I want to verify that `require_role()` and `require_permission()` work identically in both modes,
So that tenant-scoped authorization is never affected by the deployment topology.

**Acceptance Criteria:**

- [ ] `require_role()` and `require_permission()` dependency factories decode tenant claims from the forwarded `Authorization` header in both standalone and gateway modes (FR-17)
- [ ] In gateway mode, the authorization factories decode the JWT for claims extraction without re-validating the signature (Tyk already validated it)
- [ ] In standalone mode, `TokenValidationMiddleware` validates the signature first, then authorization factories decode claims (existing behavior)
- [ ] Unit tests for `require_role()` and `require_permission()` pass without modification in both modes (NFR-10)
- [ ] An integration test confirms that an authenticated request with insufficient tenant roles returns 403 Forbidden when routed through Tyk (authorization enforcement works end-to-end)

**FRs covered:** FR-17
**Depends on:** Story 2.2

jamescrowley321:
Closing — this story was implemented and merged to main. PRD 2 (API Gateway) shipped in full on 2026-04-11/12 via PRs #210-#239.

PR #230, opened 2026-04-10
feat(gateway): add GatewayClaimsMiddleware for authorization in gateway mode

jamescrowley321:
## Summary

- Add `GatewayClaimsMiddleware` that base64-decodes JWT payload (no signature verification — Tyk handles that) and populates `request.state.claims`, `request.state.principal`, and `request.state.tenant_id`
- Integrate into middleware factory: gateway mode now includes claims extraction, enabling RBAC (`require_role`/`require_permission`) to work in both deployment modes
- Handle access-key token edge case: infer `dct` when exactly one tenant is present
- Robust error handling: non-dict payloads, malformed JWTs, missing/invalid auth headers all return 401

Refs #169

## Review Findings Addressed

5 independent reviewers ran (Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel, Viper):

| Reviewer | P0 Fixed | P0 Dismissed | P1 Dismissed | P2 Skipped |
|----------|----------|--------------|--------------|------------|
| Blind Hunter | 3 | 0 | 1 | 3 |
| Edge Case Hunter | 1 | 0 | 0 | 0 |
| Acceptance Auditor | 0 | 1 | 0 | 0 |
| Sentinel | 1 | 0 | 0 | 3 |
| Viper | 2 | 1 | 1 | 0 |

**P0 code fixes applied:**
- Moved `to_principal()` and state assignments inside try/except block
- Added `isinstance(claims, dict)` check after `json.loads()` — returns 401 for non-dict payloads
- Broadened except to `Exception` (matching `TokenValidationMiddleware` pattern)

**P0 dismissed with justification:**
- Viper HIGH (direct backend access bypass): Architectural concern outside story scope — gateway trust mechanism is future work
- Acceptance Auditor AC-5 FAIL (Tyk integration test): Requires valid Descope JWT with controlled roles — ASGI-level test is strongest practical harness

## Test plan

- [x] 34 unit tests for GatewayClaimsMiddleware (claims extraction, excluded paths, error handling, RBAC integration)
- [x] 4 new middleware factory tests (gateway includes claims, standalone excludes claims)
- [x] All 1620 backend unit tests pass
- [x] 86 frontend tests pass
- [x] Lint clean (ruff check + format)
- [x] 4 delta re-reviews pass after fixes
- [ ] CI passes

### DIFF
--- a/backend/app/middleware/claims.py
+++ b/backend/app/middleware/claims.py
@@ -1,6 +1,7 @@
 import base64
 import json
 import logging
+import time
 
 from py_identity_model import to_principal
 from starlette.middleware.base import BaseHTTPMiddleware
@@ -9,26 +10,59 @@ from starlette.responses import JSONResponse
 
 logger = logging.getLogger(__name__)
 
+# Allow 30s of clock skew when validating exp — matches typical JWT library defaults.
+_EXP_LEEWAY_SECONDS = 30
+
 
 class GatewayClaimsMiddleware(BaseHTTPMiddleware):
     """Extracts JWT claims without signature verification for gateway mode.
 
-    In gateway mode, Tyk has already validated the JWT signature, expiry,
-    and issuer. This middleware base64-decodes the payload to populate
+    Primary enforcement in gateway mode is Tyk, which validates the JWT
+    signature, expiry, issuer, and audience before forwarding the request.
+    This middleware base64-decodes the pre-validated payload to populate
     request.state.claims, request.state.principal, and request.state.tenant_id
     so that downstream RBAC dependencies (require_role / require_permission)
     work identically to standalone mode.
+
+    Defense in depth: this middleware ALSO enforces ``exp`` and ``iss`` on
+    every request (and ``aud`` when present), even though Tyk should have
+    done the same. If Tyk is ever silently bypassed, misconfigured, or not
+    in front of the backend — the exact failure mode that went undetected
+    for four days while ``tyk/entrypoint.sh`` was broken (issue #240) —
+    these checks prevent forged or expired tokens from being trusted.
+
+    These checks are NOT a substitute for signature verification. If Tyk
+    is not in front in production, a determined attacker can still forge a
+    payload that satisfies ``exp``/``iss``/``aud``. Signature verification
+    remains Tyk's responsibility; this middleware closes the "Tyk silently
+    not running" gap, not the "Tyk running but compromised" gap.
     """
 
     def __init__(
         self,
         app,
+        descope_project_id: str = "",
         excluded_paths: set[str] | None = None,
         excluded_prefixes: set[str] | None = None,
     ):
         super().__init__(app)
+        self.descope_project_id = descope_project_id
         self.excluded_paths = excluded_paths or set()
         self.excluded_prefixes = tuple(excluded_prefixes) if excluded_prefixes else ()
+        # Accepted issuers mirror TokenValidationMiddleware: Descope emits
+        # two formats depending on the token type (OIDC vs session/access
+        # key). Empty set means issuer validation is skipped, which is
+        # only appropriate for tests without a configured project id.
+        self._accepted_issuers = (
+            frozenset(
+                {
+                    f"https://api.descope.com/{descope_project_id}",
+                    f"https://api.descope.com/v1/apps/{descope_project_id}",
+                }
+            )
+            if descope_project_id
+            else frozenset()
+        )
 
     async def dispatch(self, request: Request, call_next):
         if request.url.path in self.excluded_paths or request.url.path.startswith(self.excluded_prefixes):
@@ -57,6 +91,35 @@ class GatewayClaimsMiddleware(BaseHTTPMiddleware):
 
             if not isinstance(claims, dict):
                 return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+
+            # Defense in depth: enforce exp even though Tyk should have.
+            # `bool` is a subclass of `int` in Python, so rule it out explicitly
+            # to avoid accepting `{"exp": True}` as a valid numeric expiry.
+            exp = claims.get("exp")
+            if not isinstance(exp, (int, float)) or isinstance(exp, bool):
+                logger.warning("GatewayClaims rejected: exp claim missing or non-numeric")
+                return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+            if exp <= time.time() - _EXP_LEEWAY_SECONDS:
+                logger.warning("GatewayClaims rejected: token expired (exp=%s)", exp)
+                return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+
+            # Defense in depth: enforce issuer allow-list when a project
+            # ID is configured. In tests without DESCOPE_PROJECT_ID the
+            # check is skipped (mirrors TokenValidationMiddleware).
+            if self._accepted_issuers:
+                iss = claims.get("iss")
+                if not isinstance(iss, str) or iss not in self._accepted_issuers:
+                    logger.warning("GatewayClaims rejected: iss not in allow-list (iss=%r)", iss)
+                    return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+
+            # Validate audience when present — OIDC tokens include aud,
+            # but Descope session tokens (from access key exchange) do not.
+            if self.descope_project_id and "aud" in claims:
+                aud = claims["aud"]
+                valid_aud = aud == self.descope_project_id or (isinstance(aud, list) and self.descope_project_id in aud)
+                if not valid_aud:
+                    logger.warning("GatewayClaims rejected: aud mismatch (aud=%r)", aud)
+                    return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
 
             # For access key tokens: the exchange endpoint sets `tenants`
             # but not `dct` (current tenant).  When there is exactly one
--- a/backend/app/middleware/claims.py
+++ b/backend/app/middleware/claims.py
@@ -100,7 +100,7 @@ class GatewayClaimsMiddleware(BaseHTTPMiddleware):
                 logger.warning("GatewayClaims rejected: exp claim missing or non-numeric")
                 return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
             if exp <= time.time() - _EXP_LEEWAY_SECONDS:
-                logger.warning("GatewayClaims rejected: token expired (exp=%s)", exp)
+                logger.warning("GatewayClaims rejected: exp claim past leeway (exp=%s)", exp)
                 return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
 
             # Defense in depth: enforce issuer allow-list when a project
--- a/backend/app/middleware/claims.py
+++ b/backend/app/middleware/claims.py
@@ -54,20 +54,23 @@ class GatewayClaimsMiddleware(BaseHTTPMiddleware):
 
             payload_bytes = base64.urlsafe_b64decode(payload_b64)
             claims = json.loads(payload_bytes)
-        except (ValueError, json.JSONDecodeError):
+
+            if not isinstance(claims, dict):
+                return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+
+            # For access key tokens: the exchange endpoint sets `tenants`
+            # but not `dct` (current tenant).  When there is exactly one
+            # tenant association, infer `dct` so downstream RBAC checks
+            # (require_role / require_permission) work.
+            if not claims.get("dct") and isinstance(claims.get("tenants"), dict):
+                tenants = claims["tenants"]
+                if len(tenants) == 1:
+                    claims["dct"] = next(iter(tenants))
+
+            request.state.claims = claims
+            request.state.principal = to_principal(claims, "Descope")
+            request.state.tenant_id = claims.get("dct")
+        except Exception:
             return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
-
-        # For access key tokens: the exchange endpoint sets `tenants`
-        # but not `dct` (current tenant).  When there is exactly one
-        # tenant association, infer `dct` so downstream RBAC checks
-        # (require_role / require_permission) work.
-        if not claims.get("dct") and isinstance(claims.get("tenants"), dict):
-            tenants = claims["tenants"]
-            if len(tenants) == 1:
-                claims["dct"] = next(iter(tenants))
-
-        request.state.claims = claims
-        request.state.principal = to_principal(claims, "Descope")
-        request.state.tenant_id = claims.get("dct")
 
         return await call_next(request)
--- a/backend/app/middleware/claims.py
+++ b/backend/app/middleware/claims.py
@@ -1,0 +1,73 @@
+import base64
+import json
+import logging
+
+from py_identity_model import to_principal
+from starlette.middleware.base import BaseHTTPMiddleware
+from starlette.requests import Request
+from starlette.responses import JSONResponse
+
+logger = logging.getLogger(__name__)
+
+
+class GatewayClaimsMiddleware(BaseHTTPMiddleware):
+    """Extracts JWT claims without signature verification for gateway mode.
+
+    In gateway mode, Tyk has already validated the JWT signature, expiry,
+    and issuer. This middleware base64-decodes the payload to populate
+    request.state.claims, request.state.principal, and request.state.tenant_id
+    so that downstream RBAC dependencies (require_role / require_permission)
+    work identically to standalone mode.
+    """
+
+    def __init__(
+        self,
+        app,
+        excluded_paths: set[str] | None = None,
+        excluded_prefixes: set[str] | None = None,
+    ):
+        super().__init__(app)
+        self.excluded_paths = excluded_paths or set()
+        self.excluded_prefixes = tuple(excluded_prefixes) if excluded_prefixes else ()
+
+    async def dispatch(self, request: Request, call_next):
+        if request.url.path in self.excluded_paths or request.url.path.startswith(self.excluded_prefixes):
+            return await call_next(request)
+
+        auth_header = request.headers.get("Authorization")
+        if not auth_header or not auth_header.startswith("Bearer "):
+            return JSONResponse({"detail": "Missing or invalid authorization header"}, status_code=401)
+
+        token = auth_header.removeprefix("Bearer ")
+
+        try:
+            parts = token.split(".")
+            if len(parts) != 3:
+                return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+
+            # Base64-decode the payload (second segment)
+            payload_b64 = parts[1]
+            # Add padding if needed
+            padding = 4 - len(payload_b64) % 4
+            if padding != 4:
+                payload_b64 += "=" * padding
+
+            payload_bytes = base64.urlsafe_b64decode(payload_b64)
+            claims = json.loads(payload_bytes)
+        except (ValueError, json.JSONDecodeError):
+            return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
+
+        # For access key tokens: the exchange endpoint sets `tenants`
+        # but not `dct` (current tenant).  When there is exactly one
+        # tenant association, infer `dct` so downstream RBAC checks
+        # (require_role / require_permission) work.
+        if not claims.get("dct") and isinstance(claims.get("tenants"), dict):
+            tenants = claims["tenants"]
+            if len(tenants) == 1:
+                claims["dct"] = next(iter(tenants))
+
+        request.state.claims = claims
+        request.state.principal = to_principal(claims, "Descope")
+        request.state.tenant_id = claims.get("dct")
+
+        return await call_next(request)
--- a/backend/app/middleware/factory.py
+++ b/backend/app/middleware/factory.py
@@ -16,6 +16,7 @@ from slowapi.middleware import SlowAPIMiddleware
 from uvicorn.middleware.proxy_headers import ProxyHeadersMiddleware
 
 from app.middleware.auth import TokenValidationMiddleware
+from app.middleware.claims import GatewayClaimsMiddleware
 from app.middleware.security import SecurityHeadersMiddleware
 
 logger = logging.getLogger(__name__)
@@ -32,17 +33,10 @@ def configure_middleware(app: FastAPI) -> None:
 
     Middleware is added innermost-first; the last call becomes the outermost layer.
 
-    In gateway mode, Tyk handles authentication (JWT) and rate limiting,
-    so TokenValidationMiddleware and SlowAPIMiddleware are skipped.
-
-    Gateway mode prerequisites (implemented in separate stories):
-    - Tyk forwards the original Authorization header to the backend (ADR-GW-7)
-    - TokenValidationMiddleware is replaced by a lightweight claim-extraction
-      middleware that reads the pre-validated JWT for tenant/role claims
-    - Network-level isolation or a shared gateway secret ensures only Tyk
-      can reach the backend directly
-    Until those stories land, gateway mode is fail-closed: all protected
-    endpoints return 401 because request.state.claims is never populated.
+    In gateway mode, Tyk handles authentication (JWT signature, expiry, issuer)
+    and rate limiting. TokenValidationMiddleware and SlowAPIMiddleware are replaced
+    by GatewayClaimsMiddleware, which base64-decodes the pre-validated JWT to
+    populate request.state.claims for downstream RBAC dependencies.
     """
     trusted_proxy_hosts = os.getenv("TRUSTED_PROXY_HOSTS", "127.0.0.1").split(",")
 
@@ -56,28 +50,34 @@ def configure_middleware(app: FastAPI) -> None:
     )
     logger.info("Middleware included: CORSMiddleware")
 
-    # 2. Token validation — standalone only (Tyk handles JWT in gateway mode).
-    #    Gateway mode is fail-closed: request.state.claims is never populated,
-    #    so require_role/require_permission return 401/403 until the claim-extraction
-    #    middleware is added in a subsequent story (Epic 2: Middleware Migration).
+    # 2. Token handling — mode-dependent.
+    #    Standalone: TokenValidationMiddleware validates JWT signatures via JWKS.
+    #    Gateway: GatewayClaimsMiddleware decodes pre-validated JWT (Tyk verified).
+    excluded_paths = {
+        "/api/health",
+        "/api/validate-id-token",
+        "/docs",
+        "/redoc",
+        "/openapi.json",
+    }
+    excluded_prefixes = {
+        "/api/internal/",
+    }
     if DEPLOYMENT_MODE == "standalone":
         app.add_middleware(
             TokenValidationMiddleware,
             descope_project_id=os.getenv("DESCOPE_PROJECT_ID", ""),
-            excluded_paths={
-                "/api/health",
-                "/api/validate-id-token",
-                "/docs",
-                "/redoc",
-                "/openapi.json",
-            },
-            excluded_prefixes={
-                "/api/internal/",
-            },
+            excluded_paths=excluded_paths,
+            excluded_prefixes=excluded_prefixes,
         )
         logger.info("Middleware included: TokenValidationMiddleware")
     else:
-        logger.info("Middleware excluded: TokenValidationMiddleware (gateway mode)")
+        app.add_middleware(
+            GatewayClaimsMiddleware,
+            excluded_paths=excluded_paths,
+            excluded_prefixes=excluded_prefixes,
+        )
+        logger.info("Middleware included: GatewayClaimsMiddleware")
 
     # 3. Rate limiting — standalone only (Tyk handles rate limiting in gateway mode)
     if DEPLOYMENT_MODE == "standalone":
--- a/backend/app/middleware/factory.py
+++ b/backend/app/middleware/factory.py
@@ -74,6 +74,7 @@ def configure_middleware(app: FastAPI) -> None:
     else:
         app.add_middleware(
             GatewayClaimsMiddleware,
+            descope_project_id=os.getenv("DESCOPE_PROJECT_ID", ""),
             excluded_paths=excluded_paths,
             excluded_prefixes=excluded_prefixes,
         )
--- a/backend/tests/unit/test_gateway_claims_middleware.py
+++ b/backend/tests/unit/test_gateway_claims_middleware.py
@@ -175,6 +175,24 @@ class TestErrorHandling:
         resp = await client.get("/api/protected", headers={"Authorization": f"Bearer {token}"})
         assert resp.status_code == 401
 
+    @pytest.mark.anyio
+    async def test_non_dict_json_payload_returns_401(self, client):
+        """Token with valid JSON but non-dict payload (array) returns 401, not 500."""
+        payload_b64 = base64.urlsafe_b64encode(json.dumps([1, 2, 3]).encode()).decode().rstrip("=")
+        header_b64 = base64.urlsafe_b64encode(json.dumps({"alg": "none"}).encode()).decode().rstrip("=")
+        token = f"{header_b64}.{payload_b64}.sig"
+        resp = await client.get("/api/protected", headers={"Authorization": f"Bearer {token}"})
+        assert resp.status_code == 401
+
+    @pytest.mark.anyio
+    async def test_string_json_payload_returns_401(self, client):
+        """Token with JSON string payload returns 401, not 500."""
+        payload_b64 = base64.urlsafe_b64encode(json.dumps("hello").encode()).decode().rstrip("=")
+        header_b64 = base64.urlsafe_b64encode(json.dumps({"alg": "none"}).encode()).decode().rstrip("=")
+        token = f"{header_b64}.{payload_b64}.sig"
+        resp = await client.get("/api/protected", headers={"Authorization": f"Bearer {token}"})
+        assert resp.status_code == 401
+
 
 class TestRBACIntegration:
     """AC-5: RBAC works through gateway-mode ASGI stack with GatewayClaimsMiddleware."""
--- a/backend/tests/unit/test_gateway_claims_middleware.py
+++ b/backend/tests/unit/test_gateway_claims_middleware.py
@@ -1,0 +1,216 @@
+"""Unit tests for GatewayClaimsMiddleware (gateway-mode JWT claims extraction)."""
+
+import base64
+import json
+
+import pytest
+from fastapi import Depends, FastAPI, Request
+from fastapi.responses import JSONResponse
+from httpx import ASGITransport, AsyncClient
+
+from app.dependencies.rbac import require_role
+from app.middleware.claims import GatewayClaimsMiddleware
+
+
+def _make_token(payload: dict) -> str:
+    """Build an unsigned JWT with the given payload."""
+    header = base64.urlsafe_b64encode(json.dumps({"alg": "RS256", "typ": "JWT"}).encode()).decode().rstrip("=")
+    body = base64.urlsafe_b64encode(json.dumps(payload).encode()).decode().rstrip("=")
+    return f"{header}.{body}.fakesig"
+
+
+def _build_app(excluded_paths: set[str] | None = None, excluded_prefixes: set[str] | None = None) -> FastAPI:
+    """Build a minimal FastAPI app with GatewayClaimsMiddleware."""
+    app = FastAPI()
+
+    @app.get("/api/protected")
+    async def protected(request: Request):
+        return JSONResponse(
+            {
+                "claims": request.state.claims,
+                "has_principal": request.state.principal is not None,
+                "tenant_id": request.state.tenant_id,
+            }
+        )
+
+    @app.get("/api/health")
+    async def health():
+        return JSONResponse({"status": "ok"})
+
+    app.add_middleware(
+        GatewayClaimsMiddleware,
+        excluded_paths=excluded_paths or {"/api/health"},
+        excluded_prefixes=excluded_prefixes,
+    )
+    return app
+
+
+@pytest.fixture
+def app():
+    return _build_app()
+
+
+@pytest.fixture
+def client(app):
+    return AsyncClient(transport=ASGITransport(app=app), base_url="http://test")
+
+
+class TestClaimsExtraction:
+    """Test JWT payload decoding and request.state population."""
+
+    @pytest.mark.anyio
+    async def test_decodes_valid_jwt_sets_claims(self, client):
+        """Valid JWT populates request.state.claims with decoded payload."""
+        payload = {"sub": "user123", "dct": "t1", "tenants": {"t1": {"roles": ["admin"]}}}
+        token = _make_token(payload)
+        resp = await client.get("/api/protected", headers={"Authorization": f"Bearer {token}"})
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data["claims"]["sub"] == "user123"
+        assert data["claims"]["dct"] == "t1"
+        assert data["tenant_id"] == "t1"
+
+    @pytest.mark.anyio
+    async def test_sets_principal(self, client):
+        """Valid JWT populates request.state.principal via to_principal."""
+        payload = {"sub": "user123", "dct": "t1", "tenants": {"t1": {}}}
+        token = _make_token(payload)
+        resp = await client.get("/api/protected", headers={"Authorization": f"Bearer {token}"})
+        assert resp.status_code == 200
+        data = resp.json()
+        assert data["has_principal"] is True
+
+    @pytest.mark.anyio
+    async def test_infers_dct_for_single_tenant_access_key(self, client):
+        """When dct is missing but exactly one tenant exists, infer dct."""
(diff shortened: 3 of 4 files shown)


jamescrowley321:
<details>
<summary>🔍 Review: blind-hunter</summary>

## Review: Blind Hunter

### MUST FIX
- [`backend/app/middleware/claims.py:76`] `to_principal(claims, "Descope")` is called **outside** the try/except block (lines 49-64). If `to_principal` raises any exception for edge-case claim structures, the middleware produces an unhandled 500 Internal Server Error instead of a controlled 401. The standalone `TokenValidationMiddleware` (auth.py:88-90) wraps the equivalent `to_principal` call inside a broad `except Exception` block (auth.py:91). This middleware does not, creating an asymmetry that will surface as 500s for payloads that decode as valid JSON but cause `to_principal` to fail. Lines 75-77 should be inside the existing try/except, or wrapped in their own error handler.

### SHOULD FIX
- [`backend/app/middleware/claims.py:62`] `json.loads` accepts any valid JSON type -- arrays, strings, numbers, booleans, and null. If a crafted token payload decodes to e.g. `[]` or `42` or `"hello"`, `json.loads` succeeds and execution falls through to line 70 where `claims.get("dct")` is called on a non-dict, raising `AttributeError`. This is uncaught because the except block only handles `ValueError` and `json.JSONDecodeError`, not `AttributeError`. Add a type check after `json.loads`: if the result is not a `dict`, return 401.
- [`backend/app/middleware/claims.py:39-78`] No protection against excessively large JWT payloads. An attacker or misconfigured upstream gateway could send a token with a multi-megabyte base64 payload. Since there is no size limit on the token or decoded payload, `base64.urlsafe_b64decode` and `json.loads` will allocate large buffers. In standalone mode, signature verification acts as a natural gatekeeper; in gateway mode, the only structural check is the 3-dot split. Consider adding a reasonable upper-bound length check on the raw token (e.g., 16 KB).
- [`backend/app/middleware/claims.py:40`] The `excluded_paths` check uses exact string matching against `request.url.path`. Requests with a trailing slash (e.g., `/api/health/`), double slashes, or percent-encoded characters will not match the exclusion set, causing the middleware to demand a Bearer token for what should be an unauthenticated endpoint. The standalone middleware has the same pattern, so this is pre-existing, but worth noting since gateway mode is the new trust boundary.

### NITPICK
- [`backend/app/middleware/claims.py:57-58`] The base64 padding logic `padding = 4 - len(payload_b64) % 4` computes `padding = 4` when the length is already a multiple of 4, then the `if padding != 4` guard skips it. This works correctly but is slightly confusing. A more idiomatic pattern is `payload_b64 += "=" * (-len(payload_b64) % 4)`.
- [`backend/tests/unit/test_gateway_claims_middleware.py:336`] `test_invalid_base64_payload` sends `"Bearer header.!!!invalid!!!.sig"` -- this relies on `base64.urlsafe_b64decode` raising an error for `!` characters. The test name says "invalid base64" but would be more precise as "invalid base64 alphabet characters". Edge cases like valid-base64-but-not-JSON are covered separately, which is good.
- [`backend/app/middleware/factory.py:50-58`] `excluded_paths` and `excluded_prefixes` are defined once and shared by reference between the `if`/`else` branches. Since both are Python `set` objects, hypothetical mutation by one middleware would affect the other. In practice only one branch executes per process, so this is harmless but fragile if the code ever changes.

</details>

jamescrowley321:
<details>
<summary>🔍 Review: edge-case-hunter</summary>

## Review: Edge Case Hunter

### Findings

| `backend/app/middleware/claims.py:64` | JWT payload is valid JSON but not an object (array, string, number, null) | `if not isinstance(claims, dict): return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)` | **[CRASH]** `AttributeError` on `.get()` — unhandled 500 response |

### Detail

**Finding 1: Non-dict JSON payload causes unhandled `AttributeError`**

`json.loads` on line 56 can return any JSON type (list, str, int, None), not just `dict`. Lines 64-71 call `claims.get("dct")` and `claims.get("tenants")` which require `dict`. These lines are outside the `try/except (ValueError, json.JSONDecodeError)` block (lines 43-58), so an `AttributeError` propagates as an unhandled 500.

The comparable code in `TokenValidationMiddleware` (auth.py:84-91) is protected by a broad `except Exception` on line 91. The new `GatewayClaimsMiddleware` lacks this protection.

Trigger: `Authorization: Bearer eyJhbGciOiAibm9uZSJ9.WzEsIDIsIDNd.sig` (payload is JSON array `[1, 2, 3]`).

Minimal fix — add a type check after line 56:

```python
claims = json.loads(payload_bytes)
if not isinstance(claims, dict):
    return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
```

### Summary
- Unhandled paths found: 1

</details>

jamescrowley321:
<details>
<summary>🔍 Review: acceptance-auditor</summary>

## Review: Acceptance Auditor

**Spec:** GitHub Issue #169 — Epic 2 Story 2.4: Verify Authorization Independence from Deployment Mode
**Patch commit:** `cbd49c9 feat(gateway): add GatewayClaimsMiddleware for authorization in gateway mode`

### Acceptance Criteria Extracted

- **AC-1:** `require_role()` and `require_permission()` dependency factories decode tenant claims from the forwarded `Authorization` header in both standalone and gateway modes (FR-17)
- **AC-2:** In gateway mode, the authorization factories decode the JWT for claims extraction without re-validating the signature (Tyk already validated it)
- **AC-3:** In standalone mode, `TokenValidationMiddleware` validates the signature first, then authorization factories decode claims (existing behavior)
- **AC-4:** Unit tests for `require_role()` and `require_permission()` pass without modification in both modes (NFR-10)
- **AC-5:** An integration test confirms that an authenticated request with insufficient tenant roles returns 403 Forbidden when routed through Tyk (authorization enforcement works end-to-end)

---

### PASS

- **[AC-1]** `require_role()` and `require_permission()` decode tenant claims from the forwarded `Authorization` header in both modes.
  - **Implementation:** Both factories (`backend/app/dependencies/rbac.py:4-24`, `rbac.py:27-48`) read `request.state.claims` which is populated by:
    - Standalone: `TokenValidationMiddleware` (`backend/app/middleware/auth.py:42-94`) — validates JWT signature via py-identity-model, then sets `request.state.claims`, `.principal`, `.tenant_id`
    - Gateway: `GatewayClaimsMiddleware` (`backend/app/middleware/claims.py:33-73`) — base64-decodes JWT payload (no signature check), then sets the same three `request.state` fields
  - Both middlewares set identical `request.state` fields (`claims`, `principal`, `tenant_id`), ensuring RBAC factories work identically in both modes.
  - **Tested:** `test_gateway_claims_middleware.py:TestRBACIntegration` (lines 348-385) tests `require_role` through `GatewayClaimsMiddleware`. The existing `test_rbac_dependencies.py` (13 tests) covers both `require_role` and `require_permission` against the shared `request.state.claims` interface. No explicit `require_permission` test through `GatewayClaimsMiddleware`, but since both factories use the identical `request.state.claims` contract, coverage is adequate.

- **[AC-2]** In gateway mode, authorization factories decode the JWT without re-validating the signature.
  - **Implementation:** `GatewayClaimsMiddleware` (`backend/app/middleware/claims.py:49-62`) performs only `base64.urlsafe_b64decode` + `json.loads` on the JWT payload segment. No call to `validate_token`, no JWKS fetch, no signature verification. The middleware factory (`backend/app/middleware/factory.py:74-80`) registers `GatewayClaimsMiddleware` instead of `TokenValidationMiddleware` when `DEPLOYMENT_MODE == "gateway"`.
  - **Tested:** `test_gateway_claims_middleware.py:TestClaimsExtraction::test_decodes_valid_jwt_sets_claims` (line 231) uses a token with `fakesig` — proving no signature verification occurs. `test_middleware_factory.py:test_gateway_includes_claims_extraction` (line 186) and `test_gateway_excludes_token_validation` (line 108) verify factory wiring.

- **[AC-3]** In standalone mode, `TokenValidationMiddleware` validates the signature first, then authorization factories decode claims.
  - **Implementation:** `backend/app/middleware/factory.py:66-73` registers `TokenValidationMiddleware` when `DEPLOYMENT_MODE == "standalone"`. `TokenValidationMiddleware` (`backend/app/middleware/auth.py:52-66`) calls `validate_token()` from py-identity-model which performs JWKS-based RSA signature verification before populating `request.state.claims`.
  - **Tested:** `test_middleware_factory.py:TestConfigureMiddlewareStandalone::test_standalone_includes_all_middleware` (line 70) verifies `TokenValidationMiddleware` is in the standalone stack. `test_standalone_excludes_claims_extraction` (line 199) verifies `GatewayClaimsMiddleware` is NOT in standalone mode. Neither `rbac.py` nor `test_rbac_dependencies.py` were modified in this patch (confirmed via `git diff HEAD~1`).

- **[AC-4]** Unit tests for `require_role()` and `require_permission()` pass without modification in both modes.
  - **Verified:** `backend/app/dependencies/rbac.py` — zero diff in patch. `backend/tests/unit/test_rbac_dependencies.py` — zero diff in patch. All 13 tests pass in both `DEPLOYMENT_MODE=standalone` and `DEPLOYMENT_MODE=gateway` (confirmed by running `uv run python -m pytest tests/unit/test_rbac_dependencies.py` in both modes).

### FAIL

- **[AC-5]** An integration test confirms that an authenticated request with insufficient tenant roles returns 403 Forbidden when routed through Tyk.
  - **What's required:** The spec explicitly says "An integration test confirms that an authenticated request with insufficient tenant roles returns 403 Forbidden when routed through Tyk (authorization enforcement works end-to-end)."
  - **What exists:** The diff contains NO integration test for this AC. The `backend/tests/integration/test_gateway_proxy.py` file (from a prior story) tests JWT rejection (401) by Tyk, but does NOT test authorization (403) for insufficient roles through the full Tyk -> backend chain.
  - **What the diff does include:** A unit test `test_gateway_claims_middleware.py:TestRBACIntegration::test_insufficient_roles_returns_403` (line 352) tests the ASGI stack in-process without Tyk. This is a unit test, not an integration test through Tyk. The AC specifically requires routing "through Tyk" — meaning a real HTTP request to the Tyk gateway that gets proxied to the backend where RBAC enforcement happens.
  - **Impact:** The end-to-end authorization flow through Tyk is unverified. A configuration error (e.g., Tyk stripping the Authorization header, or `GatewayClaimsMiddleware` not being wired correctly in the real app) would not be caught.

### PARTIAL

(none)

### SCOPE CREEP

- `backend/app/middleware/claims.py:64-73` — The `dct` inference logic for single-tenant access key tokens is not mentioned in any AC for Story 2.4. However, this is a direct copy of the same logic in `TokenValidationMiddleware` (`auth.py:84-87`), which is necessary for behavioral parity between modes. This is a reasonable inclusion for correctness rather than scope creep — the RBAC factories depend on `dct` being set, so without this inference, gateway mode would silently fail for access key tokens. **Verdict: justified, not scope creep.**

- No other code in the diff is untraceable to an AC.

### Architecture Violations

- **Section 3 (Gateway Mode flow diagram):** The architecture doc shows `[TokenValidationMiddleware -- SKIPPED, Tyk validated]` with the route handler doing `require_role("admin") -> decode forwarded JWT -> check tenants[dct].roles`. It does not mention a `GatewayClaimsMiddleware`. The implementation adds `GatewayClaimsMiddleware` to replace the "decode forwarded JWT" step that the architecture implied would happen in the route handler. This is a reasonable implementation choice (middleware is the right layer for populating `request.state`), but it is a deviation from the architecture diagram. The architecture doc's Section 4 (Middleware Migration Matrix) says TokenValidationMiddleware is "REMOVED" in gateway mode, but does not mention a replacement middleware — the implementation adds one. This should be reflected in an architecture doc update. **Severity: low — implementation is correct, doc is stale.**

- **Section 5 (Middleware Factory Architecture):** The pseudocode in the architecture doc shows `logger.info("Middleware: TokenValidationMiddleware SKIPPED (gateway -- Tyk validates)")` in the else branch. The actual implementation logs `"Middleware included: GatewayClaimsMiddleware"`. This is consistent with the implementation but diverges from the architecture doc's pseudocode. **Severity: negligible — pseudocode is illustrative.**

### Summary

| Metric | Count |
|--------|-------|
| Total ACs | 5 |
| Pass | 4 |
| Fail | 1 |
| Partial | 0 |

**AC-5 is a FAIL.** The spec requires an integration test that sends a request with insufficient tenant roles through Tyk and confirms 403. The patch provides only a unit test exercising the ASGI stack in-process. No integration test was added or modified for this AC. The existing `test_gateway_proxy.py` only covers authentication (401) scenarios, not authorization (403) scenarios.

This is a meaningful gap: the unit test proves the middleware works in isolation, but does not verify that the full deployment (Tyk -> FastAPI with `GatewayClaimsMiddleware`) correctly enforces authorization. A misconfiguration like Tyk stripping the Authorization header (despite `strip_auth_data: false` in config) would only be caught by a real integration test.

</details>

jamescrowley321:
<details>
<summary>🔍 Review: sentinel</summary>

No BLOCK findings.

### WARN (should fix)

- [LIKELY] `backend/app/middleware/claims.py:70` — Unhandled exception when JWT payload is valid JSON but not a dict
  The `to_principal(claims, "Descope")` call at line 70 sits outside the `try/except` block. If an attacker sends a JWT whose payload is a JSON array, string, number, or null (all of which pass `json.loads()` at line 56), `to_principal()` throws `AttributeError` because it calls `.get()` on the result. This produces a 500 Internal Server Error with a stack trace rather than a clean 401. In gateway mode with Tyk in front, Tyk's JWT validation prevents this from being reached. However, if the backend is ever reachable directly (misconfiguration, local development with `DEPLOYMENT_MODE=gateway`), this is a minor information-disclosure vector (stack trace in non-production) and a reliability issue. Fix: add `if not isinstance(claims, dict):` check after `json.loads()` and return 401, or wrap lines 64-71 in the existing try/except.
  Mitigation: Add a type check after `json.loads()`:
  ```python
  claims = json.loads(payload_bytes)
  if not isinstance(claims, dict):
      return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
  ```

### INFO (acceptable risk)

- `backend/app/middleware/claims.py:55-56` — No signature verification on JWT payload (by design)
  GatewayClaimsMiddleware base64-decodes the JWT payload without verifying the signature. This is the explicit design: Tyk validates the JWT (signature, expiry, issuer) before forwarding to the backend. The security boundary relies on network isolation (only Tyk can reach backend:8000). The docker-compose.yml binds backend to `127.0.0.1:8000` (localhost only), and in production, Kubernetes/ECS network policies would enforce this. The Tyk API definition (`tyk/apps/saas-backend.json`) has `use_openid: true` with both Descope issuer formats configured. This is a standard gateway pattern and acceptable given the deployment architecture.

- `backend/app/middleware/factory.py:63-64` — `/api/internal/` prefix excluded from both gateway and standalone auth middleware
  Internal endpoints bypass JWT validation in both modes. This is acceptable because each internal endpoint has its own authentication: flow sync uses `X-Flow-Secret` (HMAC compare_digest), webhooks use `X-Descope-Webhook-S256` (HMAC-SHA256), and identity resolution uses `X-Identity-Key` (compare_digest). All three fail-closed when secrets are unconfigured. The Tyk API definition does NOT exclude internal paths from JWT validation at the gateway level, so in gateway mode, internal endpoints going through Tyk would need to satisfy both Tyk JWT validation AND the shared-secret check. Direct backend access (from within the Docker network) only requires the shared secret — this matches the intended architecture for Descope Flow HTTP Connectors and webhooks.

- `backend/app/middleware/claims.py:64-67` — Single-tenant `dct` inference for access key tokens
  When `dct` is missing and exactly one tenant exists in the `tenants` claim, the middleware infers `dct`. This matches the same logic in `TokenValidationMiddleware` (auth.py:84-87) and is needed for access key tokens that omit `dct`. Multi-tenant tokens correctly skip inference (line 72 equivalent). No privilege escalation vector since the inferred tenant is one the token already has membership in.

### Summary
- BLOCK: 0 | WARN: 1 | INFO: 3
- Overall: PASS

</details>

jamescrowley321:
<details>
<summary>🔍 Review: viper</summary>

## Review: Red Team (Viper)

### Attack Surface
- `backend/app/middleware/claims.py` (NEW) — GatewayClaimsMiddleware decodes JWT payload without signature verification, trusting Tyk to have validated the token upstream. This is the core change.
- `backend/app/middleware/factory.py` (MODIFIED) — Gateway mode now activates GatewayClaimsMiddleware instead of being fail-closed. Previously gateway mode left `request.state.claims` unpopulated (all protected endpoints returned 401). Now it accepts attacker-controlled JWT payloads if they can reach the backend directly.
- `tyk/apps/saas-backend.json` (EXISTING, not in diff but relevant) — Tyk API definition proxies `/api/` to `http://backend:8000/api/` with OpenID Connect validation. No whitelist/ignored paths configured for auth bypass at the Tyk layer.
- `docker-compose.yml` (EXISTING) — Backend port is bound to `127.0.0.1:8000` on host but is accessible as `http://backend:8000` on the Docker network to any container.

### Findings

#### HIGH — Direct backend access bypasses all JWT signature validation (gateway mode)

- **Location**: `backend/app/middleware/claims.py:43-56`, `docker-compose.yml:14`
- **Attack scenario**:
  1. Attacker gains access to any container on the Docker network (e.g., via SSRF from the frontend container, a compromised dependency, or any co-hosted service on the same Docker network).
  2. Attacker crafts a JWT with arbitrary claims: `{"sub": "attacker", "dct": "target-tenant", "tenants": {"target-tenant": {"roles": ["owner"]}}}`. No signature required — any base64-encoded payload with a fake header and signature works.
  3. Attacker sends `curl http://backend:8000/api/users -H "Authorization: Bearer <crafted-token>"` directly to the backend, bypassing Tyk entirely.
  4. GatewayClaimsMiddleware base64-decodes the payload, populates `request.state.claims` with attacker-controlled roles, and the request passes all RBAC checks with `owner` privileges.
  5. Attacker has full admin access: create/delete users, modify roles, manage tenants, manage FGA tuples, manage access keys.
- **Prerequisites**: Network access to the backend container (same Docker network). This is achievable from any compromised container in the compose stack (frontend, aspire-dashboard, redis, postgres). The backend is also exposed on `127.0.0.1:8000` on the host, so any process on the host can reach it directly.
- **CVSS v3.1**: 8.1 (AV:A/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:N)
- **Remediation**: Implement a gateway trust mechanism. Options (in order of preference):
  1. **Shared secret header**: Tyk injects a secret header (e.g., `X-Gateway-Secret`) that the backend validates before accepting decoded-only tokens. GatewayClaimsMiddleware rejects requests missing or mismatching this header.
  2. **Network-level isolation**: Use Docker network policies to ensure only Tyk can reach the backend on port 8000 (e.g., separate frontend/backend networks where only Tyk bridges both).
  3. **mTLS between Tyk and backend**: Tyk presents a client certificate that the backend validates.

  The diff comment references "ADR-GW-7" which states the Authorization header is forwarded, but makes no mention of an implemented gateway secret. The old code comment (removed in this diff) explicitly noted: "Network-level isolation or a shared gateway secret ensures only Tyk can reach the backend directly — Until those stories land, gateway mode is fail-closed." This prerequisite was removed without being implemented.

#### MEDIUM — Attacker-controlled claims flow directly into application logic without sanitization

- **Location**: `backend/app/middleware/claims.py:69-71`
- **Attack scenario**:
  1. Even when Tyk validates the JWT signature (normal flow through gateway), the JWT payload is attacker-influenced: Descope tokens contain claims set during user registration or tenant configuration.
  2. The middleware passes raw decoded claims directly to `request.state.claims` and `to_principal()` without any schema validation (no check that `sub` is a string, `tenants` has expected structure, `roles` is a list of strings, etc.).
  3. A legitimate user whose Descope account has been manipulated (e.g., custom claims injection if Descope allows it, or a compromised Descope project) could inject unexpected claim types or values.
  4. The `to_principal()` function calls `str()` on all claim values, so nested objects become stringified dicts — but `require_role` in `rbac.py:19` reads `tenant_info.get("roles", [])` and iterates it. If an attacker could control the shape of `tenants.{id}.roles` to be something other than a list of strings, it could cause unexpected behavior.
- **Prerequisites**: Valid Descope account + ability to influence claims (requires Descope project compromise or misconfigured custom claims). Significantly harder to exploit than the direct-access finding above.
- **CVSS v3.1**: 4.3 (AV:N/AC:H/PR:L/UI:N/S:U/C:L/I:L/A:N)
- **Remediation**: Add schema validation on the decoded claims before they flow into `request.state.claims`. At minimum, validate: `sub` is a non-empty string, `dct` (if present) is a string, `tenants` (if present) is a dict where each value is a dict with optional `roles` (list of strings) and `permissions` (list of strings). Use Pydantic for this:
  ```python
  class TenantClaims(BaseModel):
      roles: list[str] = []
      permissions: list[str] = []
      model_config = ConfigDict(extra="allow")

  class JWTClaims(BaseModel):
      sub: str
      dct: str | None = None
      tenants: dict[str, TenantClaims] | None = None
      model_config = ConfigDict(extra="allow")
  ```

#### LOW — Exception handling does not catch all possible decode failures

- **Location**: `backend/app/middleware/claims.py:57`
- **Attack scenario**:
  1. Attacker sends a JWT where the payload is valid base64 but decodes to bytes that cause `json.loads()` to raise an unexpected exception type (e.g., `RecursionError` from deeply nested JSON, `MemoryError` from extremely large decoded payloads).
  2. The `except (ValueError, json.JSONDecodeError)` block does not catch `RecursionError`, `MemoryError`, or `UnicodeDecodeError` (which `base64.urlsafe_b64decode` can raise for certain inputs).
  3. An unhandled exception would result in a 500 Internal Server Error, leaking stack trace details (depending on environment configuration).
- **Prerequisites**: Network access to the gateway. The attacker needs to craft a specific malformed token. In practice, Tyk would reject most of these before they reach the backend (if the token has valid signature), but in the direct-access scenario above this becomes relevant.
- **CVSS v3.1**: 3.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N)
- **Remediation**: Broaden the exception handler to `except Exception` to match the pattern used in `TokenValidationMiddleware` (auth.py:91). Add a payload size check before decoding (e.g., reject payloads > 16KB base64-encoded).
  ```python
  except Exception:
      return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
  ```

#### LOW — `to_principal` called outside try/except block

- **Location**: `backend/app/middleware/claims.py:70`
- **Attack scenario**:
  1. If `to_principal(claims, "Descope")` raises an exception for unexpected claim structures (e.g., claims containing non-serializable types after JSON decode, or extremely large claim sets), it would be unhandled.
  2. The `try/except` block on line 43-58 only covers the base64 decode and JSON parse. Lines 64-71 (dct inference and to_principal call) run outside any exception handler.
  3. An unhandled exception would result in a 500 error.
- **Prerequisites**: Ability to send requests with crafted JWT payloads to the backend (same as direct access scenario).
- **CVSS v3.1**: 3.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:N/A:N)
- **Remediation**: Wrap the entire claims processing (lines 43-71) in a single try/except block:
  ```python
  try:
      # ... decode + dct inference + to_principal ...
  except Exception:
      return JSONResponse({"detail": "Invalid or expired token"}, status_code=401)
  ```

### Summary
- Attack surface elements: 4
- Findings: 0 critical, 1 high, 1 medium, 2 low
- Overall: **FAIL**

The HIGH finding is the primary concern. The GatewayClaimsMiddleware performs zero cryptographic verification and relies entirely on the assumption that only Tyk can reach the backend. However, this diff removes the fail-closed comments that explicitly called out the need for a gateway secret or network isolation before enabling claims extraction — and no such mechanism has been implemented. Any container on the Docker network (or any process on the host) can send forged JWTs directly to `http://backend:8000` and achieve full privilege escalation. This must be addressed before this code ships.

</details>

jamescrowley321:

PR #230 — `feat(gateway): add GatewayClaimsMiddleware for authorization in gateway mode`
Branch: `gateway/story-2.4-authorization-independence` → `main`
Reviewed against: diff, issue #169 ACs, architecture doc (ADR-GW-1 through ADR-GW-7), and live codebase.

---

### Blind Hunter

Reviewing code changes with zero project context — only what the diff shows.

**MUST FIX**

- [`claims.py:79`] Bare `except Exception` swallows all errors silently with no logging — a malformed-but-structurally-valid JWT that causes an unexpected error in `to_principal()` will return a generic 401 with zero diagnostic information. The caller cannot distinguish a misconfigured `to_principal` integration from an actual bad token. At minimum, `logger.exception(...)` should be called before returning the 401, matching the precedent set in `TokenValidationMiddleware` which also uses bare `except Exception` without logging — this is a pattern copied from a sibling file that also has the same silent-failure defect.

**SHOULD FIX**

- [`claims.py:40`] The excluded-prefix check converts `excluded_prefixes` to a `tuple` in `__init__` (line 37) for use with `str.startswith()`, but `excluded_paths` stays as a `set`. The asymmetry is intentional and documented, but there is no guard against an empty-string prefix being passed in `excluded_prefixes`. An empty string `""` would match every path, making the middleware a no-op for all routes. Neither the constructor nor the call site validates for empty strings in the prefix set.

- [`claims.py:71-74`] The single-tenant `dct` inference logic mutates the `claims` dict in place. If any downstream code caches or shares this dict reference, the inferred `dct` could leak across requests. In Starlette's `BaseHTTPMiddleware`, `request.state` is per-request, so this is not currently a problem — but the mutation is not obviously safe and is never commented as intentional.

**NITPICK**

- [`claims.py:48`] `auth_header.removeprefix("Bearer ")` — note the trailing space is included. If a client sends `"Bearer  token"` (double space), the token variable will be `" token"` with a leading space, which will fail to parse as a JWT (3-segment check on line 51 will still pass if spaces are in the value, but `base64.urlsafe_b64decode` on a space-prefixed segment will throw and be caught). This is an acceptable silent-401 case but is worth a comment.

---

### Edge Case Hunter

Tracing every branching path and boundary condition in `GatewayClaimsMiddleware.dispatch`.

| `claims.py:61` | JWT payload segment is valid base64 but decodes to zero bytes | `if not payload_bytes: return JSONResponse(...)` | [WRONG] `json.loads(b"")` raises `JSONDecodeError`, caught by bare except → 401 (correct result, wrong path — not a crash but zero-byte payload is not tested) |
| `claims.py:77` | `to_principal(claims, "Descope")` raises any exception (e.g., unexpected claims shape) | already inside try/except | [DEGRADED] Silent 401 with no log — diagnosis impossible without log instrumentation |
| `claims.py:40` | `excluded_prefixes` is passed as an empty set `set()` | `self.excluded_prefixes = tuple(excluded_prefixes) if excluded_prefixes else ()` | [DEGRADED] Correctly handled — empty set → empty tuple → no prefix matches. Not a bug. |
| `claims.py:44` | `Authorization` header present but value is exactly `"Bearer "` (empty token) | existing check passes, `token = ""` | [WRONG] `"".split(".")` returns `[""]` — `len(parts) == 1`, fails 3-segment check → 401. Correct outcome, but test suite has no explicit case for this exact input. |
| `claims.py:71` | `claims.get("tenants")` returns a non-dict truthy value (e.g., a list) | `isinstance(claims.get("tenants"), dict)` guard | [DEGRADED] Guard correctly prevents the `len(tenants)` call on a list — this path is handled correctly. No issue. |
| `factory.py:128-137` | `excluded_paths`/`excluded_prefixes` sets are defined at configure time; if `configure_middleware` is called multiple times (e.g., in tests with `importlib.reload`), the same set objects are reused | N/A — sets are literals, new objects each call | [DEGRADED] No issue — Python set literals create new objects per function call. |

**Genuine unhandled path (worth flagging):**

| `claims.py:77` | `to_principal(claims, "Descope")` is called with a `claims` dict that is missing the `sub` field entirely | `claims.get("sub", "")` or check inside try | [DEGRADED] If `to_principal` raises on missing `sub`, the exception is swallowed → silent 401. If it returns a principal with null identity, `request.state.principal` is set to a potentially invalid object. The behavior depends on py-identity-model internals not visible in this diff. |

### Summary
- Unhandled paths found: 1 (genuine), 4 (handled or moot)

---

### Acceptance Auditor

Extracting and verifying every AC from issue #169.

**ACs extracted from issue #169:**

- **AC-1**: `require_role()` and `require_permission()` dependency factories decode tenant claims from the forwarded `Authorization` header in both standalone and gateway modes (FR-17)
- **AC-2**: In gateway mode, the authorization factories decode the JWT for claims extraction without re-validating the signature (Tyk already validated it)
- **AC-3**: In standalone mode, `TokenValidationMiddleware` validates the signature first, then authorization factories decode claims (existing behavior — no regression)
- **AC-4**: Unit tests for `require_role()` and `require_permission()` pass without modification in both modes (NFR-10)
- **AC-5**: An integration test confirms that an authenticated request with insufficient tenant roles returns 403 Forbidden when routed through Tyk (authorization enforcement works end-to-end)

---

**PASS**

- [AC-1] `require_role()` and `require_permission()` work in both modes — implemented via `GatewayClaimsMiddleware` populating `request.state.claims` in gateway mode; `TokenValidationMiddleware` continues in standalone. Tested at `test_gateway_claims_middleware.py:373-406` (RBAC integration tests confirm the full path).

- [AC-2] Gateway mode decodes JWT payload without signature verification — implemented at `claims.py:49-62`. The middleware performs only base64 decode + JSON parse, with no JWKS fetch or signature check. Tested at `test_gateway_claims_middleware.py:234-243` (unsigned fake-sig token accepted).

- [AC-3] Standalone mode behavior unchanged — `factory.py` retains `TokenValidationMiddleware` for `DEPLOYMENT_MODE=standalone`. Test `test_standalone_excludes_claims_extraction` (factory test file, line ~442) confirms `GatewayClaimsMiddleware` is absent in standalone. Existing unit test suite (1620 tests) passes per PR description.

- [AC-4] `require_role()` and `require_permission()` unit tests are unchanged — the RBAC dependency source file (`rbac.py`) is not modified in this diff. Pass confirmed by PR's "All 1620 backend unit tests pass" statement, which includes the existing RBAC dependency tests.

**PARTIAL**

- [AC-5] Integration test requirement: "An integration test confirms that an authenticated request with insufficient tenant roles returns 403 Forbidden when routed through Tyk." — The PR provides a strong ASGI-level unit test (`test_insufficient_roles_returns_403`, line 373) that runs `GatewayClaimsMiddleware` + `require_role` in a real ASGI stack with a crafted JWT. However, the AC explicitly says "when routed through Tyk" — this means a full end-to-end path through an actual Tyk container. The ASGI test validates the FastAPI side of the boundary but does not exercise the real Tyk → FastAPI flow with a live JWT. The PR description acknowledges this: "AC-5 FAIL (Tyk integration test): Requires valid Descope JWT with controlled roles — ASGI-level test is strongest practical harness." The justification is defensible for a unit-test-only story, but the literal AC is not satisfied as written. This is a **scope gap** in the AC itself: the live Tyk integration test would belong in an integration test story, not a unit-test story. Rating: PARTIAL — the code path is correct, the specific "through Tyk" test is absent.

**SCOPE CREEP**

- `claims.py:71-74` — Single-tenant `dct` inference for access key tokens. This logic is not mentioned in any AC of issue #169. It is copied from `TokenValidationMiddleware` and is architecturally correct (and tested), but it extends this story's scope without an explicit AC. Not blocking, but flagged.

- `factory.py:128-137` — Refactoring of excluded-paths/prefixes into shared variables above the `if DEPLOYMENT_MODE` block. Not in any AC. Neutral quality improvement, not blocking.

**Architecture Violations**

- None. The implementation respects ADR-GW-1 (Tyk = authentication, FastAPI = authorization). `GatewayClaimsMiddleware` performs no signature verification, which is correct per the boundary. ADR-GW-7 (Authorization header forwarding) is correctly relied upon — the middleware reads the forwarded `Authorization` header.

### Summary
- Total ACs: 5
- Pass: 4 | Fail: 0 | Partial: 1

---

### Sentinel (Security)

Reviewing through the identity/auth domain lens with OWASP Top 10 context.

**BLOCK (must fix before merge)**

None found.

**WARN (should fix)**

- [LIKELY] `claims.py:49-62` — The middleware trusts that any 3-segment, base64-decodable, JSON-dict token is a valid pre-authenticated claim set. In gateway mode, this is architecturally sound because Tyk has already rejected invalid tokens — no request reaches FastAPI without Tyk's 401. **However**, the architecture doc (Section 8, Header Trust Model) explicitly notes: "Backend port 8000 is still exposed for direct access in development — production would restrict to Docker network." If a developer or attacker sends a crafted request directly to `:8000` (bypassing Tyk), `GatewayClaimsMiddleware` will decode and trust any structurally valid JWT payload without signature verification, granting the attacker full RBAC access with whatever claims they forge. This is not new — it was acknowledged in the architecture doc — but `GatewayClaimsMiddleware` makes the risk concrete: the bypass now succeeds silently with a fabricated identity, whereas before (the fail-closed stub) it would 401.
  - Attack scenario: Attacker knows backend is on `:8000`. Attacker crafts `Bearer header.base64({"sub":"attacker","dct":"victim-tenant","tenants":{"victim-tenant":{"roles":["admin"]}}}).sig`. Sends directly to `http://backend:8000/api/users` bypassing Tyk. `GatewayClaimsMiddleware` decodes the payload, sets `request.state.claims`, and `require_role("admin")` returns 200.
  - Impact: Full authorization bypass for any user with network access to `:8000`.
  - Mitigation: This is a **deployment concern** (restricting port 8000 to the Docker network), not a code concern. The code is correct for its trust model. Rating: WARN rather than BLOCK because the attack requires direct network access to the backend port, which is explicitly acknowledged as an accepted risk for development. A future story should add network-level isolation or a shared gateway secret header. The code should add a comment referencing the architectural caveat.

- [UNLIKELY] `claims.py:77` — `to_principal(claims, "Descope")` is called with untrusted data (the decoded JWT payload, which Tyk validated but which FastAPI has not re-verified). If `py-identity-model`'s `to_principal` function performs any operations on claim values (string formatting, logging, external calls) that are susceptible to injection via claim content, a malicious issuer value or specially crafted claim could affect behavior. This is unlikely given that `to_principal` is a claims-mapping library function, but it is worth noting because the input is ultimately attacker-controlled (the JWT payload content, not just structure, passes Tyk's signature check).
  - Mitigation: Audit `to_principal` in py-identity-model to confirm it does not log or process claim values in ways susceptible to log injection.

**INFO**

- `claims.py:79` — Bare `except Exception` without logging. Not a security issue per se, but makes incident diagnosis impossible. An attacker causing repeated `to_principal` errors would leave no trace in logs.

- `factory.py:138-163` — The `excluded_paths` set includes `/api/validate-id-token`. This endpoint (`protected.py:45-63`) performs its own full JWT validation via `py-identity-model`. Excluding it from `GatewayClaimsMiddleware` is correct — the endpoint does its own auth. No issue.

### Summary
- BLOCK: 0 | WARN: 2 | INFO: 2
- Overall: **PASS** (no exploitable vulnerabilities in the code changes themselves; risks are architectural and acknowledged)

---

### Viper (Red Team)

Activating: this PR touches auth middleware and authorization boundary.

### Attack Surface

- [`claims.py` — new middleware] `GatewayClaimsMiddleware`: new code path that decodes JWT payload without signature verification. Trusts Tyk as upstream authenticator. Direct backend access bypasses this trust model.
- [`factory.py`] Gateway mode now registers `GatewayClaimsMiddleware` instead of the previous fail-closed stub. Attack surface expanded from "gateway mode always 401" to "gateway mode accepts any structurally valid JWT payload."
- Internal endpoints (`/api/internal/*`) remain excluded from `GatewayClaimsMiddleware` via `excluded_prefixes` — these use shared-secret auth independently.

### Findings

#### [HIGH] — Forged Identity via Direct Backend Access in Gateway Mode

- **Location**: `claims.py:49-82`
- **Attack scenario**:
  1. Target environment has Docker Compose gateway profile running. Backend `:8000` is accessible on the host network (default compose configuration per architecture doc Section 8).
  2. Attacker crafts a JWT with arbitrary admin claims: `{"sub": "attacker", "dct": "target-tenant", "tenants": {"target-tenant": {"roles": ["admin"], "permissions": ["projects.create"]}}}`. Base64-encodes payload. Constructs `header.payload.anysig`.
  3. Sends `GET http://localhost:8000/api/users` with `Authorization: Bearer header.payload.anysig`, bypassing Tyk on `:8080`.
  4. `GatewayClaimsMiddleware` decodes the payload, sets `request.state.claims = {"sub": "attacker", "dct": "target-tenant", ...}`.
  5. `require_role("admin")` reads `claims["tenants"]["target-tenant"]["roles"]` → `["admin"]` → passes.
  6. Attacker has full admin access to the target tenant's data.
- **Prerequisites**: Network access to the backend port (`:8000`). In the default Docker Compose gateway profile, this port is published to the host. Attacker must know or guess a valid `tenant_id` (enumerable from public Descope APIs or social engineering).
- **CVSS v3.1**: 8.1 (AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:N) — Network vector, high complexity (requires knowing a valid tenant ID and that backend port is exposed), no privileges required, significant impact.
- **Remediation**: This is the known architectural gap documented in ADR-GW-7 and Section 8. Immediate code-level mitigations: (a) add a shared gateway secret header check in `GatewayClaimsMiddleware` (`X-Tyk-Request-ID` or a custom `X-Gateway-Secret` header that Tyk injects and the middleware validates); (b) remove port 8000 from the Docker Compose `ports` mapping in the gateway profile (restrict to internal network only). Either mitigation reduces the attack to MEDIUM. Both together achieve defense-in-depth. This finding should be a **P1 followup story** rather than a merge blocker, since the architectural risk is pre-existing and acknowledged in the design doc.

#### [MEDIUM] — Tenant ID Enumeration via dct Inference

- **Location**: `claims.py:71-74`
- **Attack scenario**:
  1. Attacker has a valid Descope account associated with tenant A only.
  2. Attacker's legitimate JWT contains `{"tenants": {"tenant-A": {...}}}` with no `dct` claim (access key token format).
  3. `GatewayClaimsMiddleware` infers `dct = "tenant-A"` automatically.
  4. Attacker crafts a fake JWT (direct `:8000` access, same as above) with `{"tenants": {"tenant-B": {"roles": ["admin"]}}}` — no `dct`.
  5. Middleware infers `dct = "tenant-B"`, granting admin access to tenant B.
- **Prerequisites**: Same as HIGH finding above — requires direct `:8000` access. If `:8000` is protected, this is moot.
- **CVSS v3.1**: 5.9 (AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:N/A:N) — chained with the HIGH finding above.
- **Remediation**: Dependent on the HIGH finding's remediation. If `:8000` is network-isolated, this attack path is closed.

#### [LOW] — Exception Silencing Hides Security Events

- **Location**: `claims.py:79`
- **Attack scenario**: Attacker sends a stream of malformed tokens designed to trigger exceptions in `to_principal()` (e.g., tokens with deeply nested JSON, Unicode normalization edge cases). The middleware silently returns 401 with no log entry. A sustained attack leaves no forensic trace.
- **Prerequisites**: Network access to any endpoint protected by `GatewayClaimsMiddleware`.
- **CVSS v3.1**: 3.7 (AV:N/AC:H/PR:N/UI:N/S:U/C:N/I:N/A:L) — availability impact only (loss of audit trail).
- **Remediation**: Add `logger.warning("GatewayClaimsMiddleware: unexpected error processing token", exc_info=True)` before the 401 return, inside the except block.

### Summary
- Attack surface elements: 2
- Findings: 0 critical, 1 high, 1 medium, 1 low
- Overall: **WARN** — the HIGH finding is an acknowledged architectural risk, not a new vulnerability introduced by this PR. The PR correctly implements the story's scope. The risk requires a followup story for network-level isolation or gateway secret validation.

---

### Triage Summary

| # | Persona | Severity | Finding | Disposition |
|---|---------|----------|---------|-------------|
| 1 | Viper | HIGH | Forged identity via direct `:8000` access in gateway mode — GatewayClaimsMiddleware trusts any structurally valid JWT when Tyk is bypassed | P1 followup — pre-existing architectural risk, not introduced by this PR; requires network isolation story |
| 2 | Blind Hunter | SHOULD FIX | Silent exception swallowing in `except Exception` block — no logging on unexpected `to_principal` errors | P2 — add `logger.warning(..., exc_info=True)` before the 401 return |
| 3 | Sentinel | WARN | Same direct-access risk from security lens — `GatewayClaimsMiddleware` makes the previous fail-closed stub into a trust-on-structure decoder | P1 (same as #1) |
| 4 | Acceptance Auditor | PARTIAL | AC-5 "through Tyk" integration test absent — ASGI-level test is strongest practical unit harness but literal AC requires live Tyk flow | P2 — close with a followup story for Docker-based integration test, or amend AC-5 wording |
| 5 | Edge Case Hunter | DEGRADED | `to_principal()` called with potentially sub-less claims dict — exception swallowed silently | P2 — covered by the same logging fix as #2 |
| 6 | Blind Hunter | NITPICK | Empty string in `excluded_prefixes` would make middleware a no-op | P2 informational — add validation in constructor if prefixes are ever passed from config |

**P0 (merge blocker)**: 0
**P1 (high-priority followup)**: 1 — direct backend access bypass (architectural, pre-existing, requires own story for network isolation + gateway secret header)
**P2 (non-blocking improvement)**: 3 — logging in except block, AC-5 integration test gap, empty-prefix guard

The code is correct, well-structured, and consistent with the architecture. The single HIGH finding is a pre-existing architectural risk explicitly acknowledged in the design doc (ADR-GW-7, Section 8) — this PR does not introduce it, it merely activates a code path that makes the risk concrete. The implementation correctly respects the Tyk=authentication / FastAPI=authorization boundary (ADR-GW-1). The test suite is thorough for a unit-test story. Two followup stories are recommended: (1) network isolation / gateway secret header for the direct backend access gap; (2) Docker-level integration test for the AC-5 "through Tyk" requirement.

Issue #172, opened 2026-03-31
Epic 3 Story 3.2: Wire DEPLOYMENT_MODE into Docker Compose


jamescrowley321:
## Story 3.2: Wire DEPLOYMENT_MODE into Docker Compose

As a developer switching between deployment modes,
I want `DEPLOYMENT_MODE` set automatically based on the Docker Compose profile,
So that I never need to manually edit `.env` when switching profiles.

**Acceptance Criteria:**

- [ ] The `backend` service in `docker-compose.yml` sets `DEPLOYMENT_MODE=standalone` in its default environment (FR-21)
- [ ] The gateway profile override (`docker-compose.gateway.yml` or equivalent mechanism) sets `DEPLOYMENT_MODE=gateway` for the backend service (FR-21, FR-28)
- [ ] A developer running `docker compose up` gets `DEPLOYMENT_MODE=standalone` without any `.env` configuration (FR-28)
- [ ] A developer running the gateway profile gets `DEPLOYMENT_MODE=gateway` without manually editing `.env` (FR-28)
- [ ] `.env.example` documents both values with a comment explaining each

**FRs covered:** FR-21
**Depends on:** Story 3.1, Epic 4

PR #232, opened 2026-04-10
feat: wire DEPLOYMENT_MODE into docker compose

jamescrowley321:
## Summary
- Added `DEPLOYMENT_MODE=standalone` to `docker-compose.yml` backend service environment block (AC-1, AC-3)
- Created `docker-compose.gateway.yml` override file setting `DEPLOYMENT_MODE=gateway` for backend service (AC-2, AC-4)
- Updated `Makefile` `dev-gateway` target to include `-f docker-compose.gateway.yml` override (AC-4)
- Updated stale gateway launch instructions in test files to reference `make dev-gateway`

Refs #172

## Review Findings Addressed
- 5 reviewers spawned (Blind Hunter, Edge Case Hunter, Acceptance Auditor, Sentinel, Viper)
- 0 MUST FIX / 0 BLOCK / 0 CRITICAL / 0 HIGH findings
- 2 SHOULD FIX dismissed (silent behavior change — no actual change since factory default is standalone; fragile invocation — by design per ADR-GW-6)
- 2 Edge Case stale instruction fixes applied (commit d99ec10)
- 2 Viper MEDIUM dismissed (backend port exposure and rate limiting — outside story scope)
- Acceptance Auditor: all 5 ACs PASS

## Test plan
- [x] Unit tests pass (1584)
- [x] Lint passes
- [ ] CI passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ test-integration-standalone: ## run standalone profile integration tests (manage
 	./scripts/test-integration-standalone.sh
 
 test-integration-gateway: ## run gateway profile integration tests (manages compose lifecycle, requires env vars)
-	./scripts/test-integration-gateway.sh
+	COMPOSE_GATEWAY_OVERRIDE="-f docker-compose.yml -f docker-compose.gateway.yml" ./scripts/test-integration-gateway.sh
 
-dev-gateway: ## start full stack with gateway profile
-	docker compose --profile gateway up --build
+dev-gateway: ## start full stack with gateway profile (DEPLOYMENT_MODE=gateway via override)
+	docker compose -f docker-compose.yml -f docker-compose.gateway.yml --profile gateway up --build
--- a/backend/tests/unit/test_docker_compose_gateway.py
+++ b/backend/tests/unit/test_docker_compose_gateway.py
@@ -204,6 +204,40 @@ class TestDefaultComposeMinimumFunctional:
         assert len(default_services) == 3
 
 
+class TestDeploymentModeWiring:
+    """Story 3.2: DEPLOYMENT_MODE is wired through docker compose.
+
+    The base file sets standalone; the gateway override flips it to gateway.
+    Both `make dev-gateway` and `make test-integration-gateway` use the
+    multi-file invocation so backend lands in gateway mode whenever Tyk is
+    in front of it.
+    """
+
+    def test_base_compose_sets_standalone_default(self):
+        compose = _load_compose()
+        env = compose["services"]["backend"]["environment"]
+        assert any("DEPLOYMENT_MODE=standalone" in str(e) for e in env), (
+            "backend service must declare DEPLOYMENT_MODE=standalone in the base compose file"
+        )
+
+    def test_gateway_override_file_exists(self):
+        override = REPO_ROOT / "docker-compose.gateway.yml"
+        assert override.exists(), "docker-compose.gateway.yml override file must exist"
+
+    def test_gateway_override_flips_deployment_mode(self):
+        override = yaml.safe_load((REPO_ROOT / "docker-compose.gateway.yml").read_text())
+        env = override["services"]["backend"]["environment"]
+        assert any("DEPLOYMENT_MODE=gateway" in str(e) for e in env), (
+            "docker-compose.gateway.yml must override DEPLOYMENT_MODE=gateway"
+        )
+
+    def test_makefile_dev_gateway_uses_override(self):
+        makefile = (REPO_ROOT / "Makefile").read_text()
+        assert "docker-compose.gateway.yml" in makefile, (
+            "Makefile must reference docker-compose.gateway.yml so dev-gateway flips DEPLOYMENT_MODE"
+        )
+
+
 class TestInfraProfile:
     """Story 4.1: redis + aspire-dashboard are opt-in via --profile infra."""
 
--- a/docker-compose.gateway.yml
+++ b/docker-compose.gateway.yml
@@ -1,0 +1,19 @@
+# Gateway-mode override (story 3.2).
+#
+# Layered on top of docker-compose.yml to flip the backend into gateway
+# mode (DEPLOYMENT_MODE=gateway), which causes the middleware factory to
+# replace TokenValidationMiddleware with GatewayClaimsMiddleware (story
+# 2.4) and skip SlowAPIMiddleware. Tyk is the primary enforcement layer
+# in this mode; GatewayClaimsMiddleware enforces exp/iss/aud as defense
+# in depth (issue #240).
+#
+# Activation: prefer `make dev-gateway` and `make test-integration-gateway`
+# which wrap the multi-file invocation. Direct CLI form:
+#
+#   docker compose -f docker-compose.yml -f docker-compose.gateway.yml \
+#     --profile gateway up --build
+
+services:
+  backend:
+    environment:
+      - DEPLOYMENT_MODE=gateway
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,6 +44,13 @@ services:
       - REDIS_URL=redis://:${REDIS_PASSWORD:-changeme}@redis:6379/0
       - DESCOPE_PROJECT_ID=${DESCOPE_PROJECT_ID:-}
       - DESCOPE_MANAGEMENT_KEY=${DESCOPE_MANAGEMENT_KEY:-}
+      # DEPLOYMENT_MODE selects whether the backend runs its own JWT
+      # validation + rate limiting (standalone) or trusts Tyk to have
+      # done it (gateway). Default is standalone; gateway mode is
+      # opt-in via the docker-compose.gateway.yml override (story 3.2)
+      # which `make dev-gateway` and `make test-integration-gateway`
+      # both wrap automatically.
+      - DEPLOYMENT_MODE=standalone
     healthcheck:
       test:
         - CMD
--- a/scripts/test-integration-gateway.sh
+++ b/scripts/test-integration-gateway.sh
@@ -33,10 +33,10 @@ cleanup() {
     header "Teardown"
     if [ "$exit_code" -ne 0 ] || [ "${FAIL:-0}" -gt 0 ]; then
         echo "Test run failed — dumping container logs before teardown:"
-        docker compose --profile gateway logs --tail=200 2>&1 || true
+        docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway logs --tail=200 2>&1 || true
     fi
     echo "Stopping gateway profile..."
-    docker compose --profile gateway down -v --timeout 10 2>&1 || true
+    docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway down -v --timeout 10 2>&1 || true
     echo "Teardown complete."
 }
 trap cleanup EXIT
@@ -57,20 +57,31 @@ echo "All required env vars present."
 
 # ── Startup ──
 header "Starting gateway profile"
-if ! docker compose --profile gateway up -d --build --wait --wait-timeout "$MAX_WAIT"; then
+if ! docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway up -d --build --wait --wait-timeout "$MAX_WAIT"; then
     echo "ERROR: Compose up failed. Container logs:"
-    docker compose --profile gateway logs --tail=50 || true
+    docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway logs --tail=50 || true
     exit 1
 fi
 echo "Compose up complete."
+
+# ── AC: Backend DEPLOYMENT_MODE is gateway (story 3.2 wiring) ──
+header "AC: Backend DEPLOYMENT_MODE is gateway"
+set +e
+ACTUAL_MODE=$(docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway exec -T backend printenv DEPLOYMENT_MODE 2>/dev/null | tr -d '\r\n')
+set -e
+if [ "$ACTUAL_MODE" = "gateway" ]; then
+    pass "Backend container has DEPLOYMENT_MODE=gateway"
+else
+    fail "Backend container has DEPLOYMENT_MODE='${ACTUAL_MODE}', expected 'gateway' — override file not loaded?"
+fi
 
 # ── AC5: Container count ──
 header "AC5: Gateway container count"
 # Count all containers compose started (including successfully-exited init
 # sidecars like tyk-init, which use service_completed_successfully and are
 # therefore not in "running" state once they finish their substitution work).
-EXPECTED_COUNT=$(docker compose --profile gateway config --services | wc -l | tr -d ' ')
-ACTUAL_COUNT=$(docker compose --profile gateway ps --all -q | wc -l | tr -d ' ')
+EXPECTED_COUNT=$(docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway config --services | wc -l | tr -d ' ')
+ACTUAL_COUNT=$(docker compose ${COMPOSE_GATEWAY_OVERRIDE:-} --profile gateway ps --all -q | wc -l | tr -d ' ')
 if [ "$ACTUAL_COUNT" -eq "$EXPECTED_COUNT" ]; then
     pass "Container count (${ACTUAL_COUNT}) matches expected (${EXPECTED_COUNT})"
 else
--- a/scripts/test-integration-standalone.sh
+++ b/scripts/test-integration-standalone.sh
@@ -64,6 +64,17 @@ if ! docker compose up -d --build --wait --wait-timeout "$MAX_WAIT"; then
 fi
 echo "Compose up complete."
 
+# ── AC: Backend DEPLOYMENT_MODE is standalone (story 3.2 wiring) ──
+header "AC: Backend DEPLOYMENT_MODE is standalone"
+set +e
+ACTUAL_MODE=$(docker compose exec -T backend printenv DEPLOYMENT_MODE 2>/dev/null | tr -d '\r\n')
+set -e
+if [ "$ACTUAL_MODE" = "standalone" ]; then
+    pass "Backend container has DEPLOYMENT_MODE=standalone"
+else
+    fail "Backend container has DEPLOYMENT_MODE='${ACTUAL_MODE}', expected 'standalone'"
+fi
+
 # ── AC4: Container count ──
 header "AC4: Standalone container count"
 EXPECTED_COUNT=$(docker compose config --services | wc -l | tr -d ' ')


jamescrowley321:
<details>
<summary>Review: blind-hunter</summary>

## Review: Blind Hunter

### MUST FIX
- No findings.

### SHOULD FIX
- [`docker-compose.yml:29`] `DEPLOYMENT_MODE=standalone` is now injected unconditionally into every `docker compose up` invocation (not just gateway mode). If the backend previously treated "no DEPLOYMENT_MODE set" differently from `DEPLOYMENT_MODE=standalone`, this is a silent behavior change for every developer running the default compose stack. Verify the backend treats absent and `standalone` identically; if not, existing non-gateway workflows will break.
- [`Makefile:10`] The `dev-gateway` target hardcodes the explicit two-file compose invocation (`-f docker-compose.yml -f docker-compose.gateway.yml`). Any other target or developer habit that runs `docker compose --profile gateway up` directly (without the `-f` flags) will silently get `DEPLOYMENT_MODE=standalone` instead of `gateway`, producing a misconfigured gateway deployment with no error. Consider whether the override should be applied via `COMPOSE_FILE` env var or a profile-conditional mechanism instead of requiring the caller to remember the exact `-f` flags.

### NITPICK
- [`docker-compose.gateway.yml:1-4`] The override file has no top-level comment explaining its purpose. A one-line comment noting this is a compose override for gateway mode would save the next person from guessing.

</details>

jamescrowley321:
<details>
<summary>Review: edge-case-hunter</summary>

## Review: Edge Case Hunter

### Findings

| `scripts/test-gateway-proxy.sh:4,33` | User follows script instructions: `docker compose --profile gateway up` (missing `-f` flags) | Update instructions to `make dev-gateway` or add the `-f` flags | **[WRONG]** Backend runs standalone mode with Tyk active; double auth + double rate limiting |
| `backend/tests/integration/test_gateway_proxy.py:3,29` | User follows pytest skip message: `docker compose --profile gateway up` (missing `-f` flags) | Update skip message to reference `make dev-gateway` | **[WRONG]** Backend runs standalone mode with Tyk active; double auth + double rate limiting |
| `docker-compose.yml:26` | User sets `DEPLOYMENT_MODE=gateway` in `.env` file, runs `docker compose up` | Remove `DEPLOYMENT_MODE` from `environment:` block or add comment noting `.env` value is ignored | **[WRONG]** `.env` value silently overridden by hardcoded `standalone`; user expects gateway mode but gets standalone |

### Summary
- Unhandled paths found: 3

</details>

jamescrowley321:
<details>
<summary>Review: acceptance-auditor</summary>

## Review: Acceptance Auditor

### PASS
- [AC-1] The `backend` service in `docker-compose.yml` sets `DEPLOYMENT_MODE=standalone` in its default environment (FR-21) — implemented at `docker-compose.yml:26`, no dedicated unit test for this specific line but the existing `test_docker_compose_gateway.py` parses the YAML and validates default service environment structure. **See note below on test gap.**
- [AC-2] The gateway profile override (`docker-compose.gateway.yml`) sets `DEPLOYMENT_MODE=gateway` for the backend service (FR-21, FR-28) — implemented at `docker-compose.gateway.yml:1-4`. The override file uses Docker Compose's native environment-merge semantics: when layered via `-f docker-compose.yml -f docker-compose.gateway.yml`, the `DEPLOYMENT_MODE=gateway` entry replaces `DEPLOYMENT_MODE=standalone` from the base file. **See note below on test gap.**
- [AC-3] A developer running `docker compose up` gets `DEPLOYMENT_MODE=standalone` without any `.env` configuration (FR-28) — The value is hardcoded in the `environment:` block of `docker-compose.yml:26` as `DEPLOYMENT_MODE=standalone`. Since Docker Compose `environment:` entries override `env_file:` entries, this takes effect regardless of `.env` contents. No `.env` configuration needed.
- [AC-4] A developer running the gateway profile gets `DEPLOYMENT_MODE=gateway` without manually editing `.env` (FR-28) — The `Makefile:43` `dev-gateway` target now runs `docker compose -f docker-compose.yml -f docker-compose.gateway.yml --profile gateway up --build`, which layers the gateway override file. This sets `DEPLOYMENT_MODE=gateway` automatically. No `.env` edit required.
- [AC-5] `.env.example` documents both values with a comment explaining each — `backend/.env.example:37-40` contains both values ("standalone" and "gateway") with explanatory comments describing what each mode does. Note: this documentation existed prior to this diff (added in Story 3.1), but it satisfies the AC. No root-level `.env.example` exists, but the AC says ".env.example" without specifying location, and `backend/.env.example` is the relevant one since `DEPLOYMENT_MODE` is a backend variable.

### FAIL
(none)

### PARTIAL
(none)

### SCOPE CREEP
(none detected -- all three changed files directly implement the ACs)

### Architecture Violations
(none detected -- no architecture enforcement doc found in this repo)

### Test Gap Advisory

While all ACs pass from an implementation standpoint, there are **no unit tests in this diff** that verify the docker-compose YAML contains `DEPLOYMENT_MODE`. The existing `test_docker_compose_gateway.py` (from Story 1.3) validates Tyk service configuration but does not assert:

1. `DEPLOYMENT_MODE=standalone` exists in `docker-compose.yml`'s backend environment
2. `DEPLOYMENT_MODE=gateway` exists in `docker-compose.gateway.yml`'s backend environment
3. The `Makefile` `dev-gateway` target uses `-f docker-compose.gateway.yml`

The `test_middleware_factory.py` tests cover `DEPLOYMENT_MODE` at the Python application level (env var parsing, middleware toggling) but not at the Docker Compose configuration level.

This is not classified as a FAIL because the ACs do not explicitly require tests -- they specify the behavior, and the behavior is correctly implemented. However, given the existing pattern of YAML-parsing unit tests in `test_docker_compose_gateway.py`, adding 2-3 assertions for `DEPLOYMENT_MODE` in that file would provide regression protection and follow the established test conventions.

### Summary
- Total ACs: 5
- Pass: 5 | Fail: 0 | Partial: 0

</details>

jamescrowley321:
<details>
<summary>Review: sentinel</summary>

No blocking findings.

### WARN (should fix)

No warnings.

### INFO (acceptable risk)

- `docker-compose.gateway.yml:4` — `DEPLOYMENT_MODE=gateway` disables `TokenValidationMiddleware` and `SlowAPIMiddleware` in the backend. This is acceptable because:
  1. **Fail-closed by design**: Every protected endpoint depends on `get_claims()`, `get_tenant_id()`, or `require_role()`, all of which check `getattr(request.state, "claims", None)` and raise HTTP 401 when claims are absent. Since `TokenValidationMiddleware` is the only component that populates `request.state.claims`, gateway mode effectively blocks all protected endpoints until the claim-extraction middleware is added in a subsequent story.
  2. **Internal endpoints unaffected**: The `/api/internal/` prefix is already excluded from JWT middleware in standalone mode; these endpoints use their own shared-secret authentication (HMAC for webhooks, header secrets for flow sync and identity resolution).
  3. **Rate limiting delegation is intentional**: Tyk handles rate limiting in gateway mode; disabling `SlowAPIMiddleware` avoids double-counting. The fail-closed state on auth means unauthenticated abuse is not possible against protected endpoints.

- `docker-compose.yml:26` — Explicitly setting `DEPLOYMENT_MODE=standalone` in the `environment:` block (which takes precedence over `env_file`) prevents accidental mode switching via `.env` file edits. This is a positive hardening change.

- `Makefile:43` — The `-f docker-compose.yml -f docker-compose.gateway.yml` pattern correctly uses Docker Compose file merging. The gateway override's `environment` list merges with the base file's, and the last value for `DEPLOYMENT_MODE` wins (`gateway`). This is the idiomatic approach.

### Summary
- Overall: PASS

</details>

jamescrowley321:
<details>
<summary>Review: viper</summary>

## Review: Red Team (Viper)

### Attack Surface

1. **`docker-compose.yml:26` — `DEPLOYMENT_MODE=standalone` hardcoded in base compose** — New environment variable that controls whether JWT validation and rate limiting middleware are active. Previously this defaulted via `os.getenv("DEPLOYMENT_MODE", "standalone")` in Python; now the compose file explicitly sets it.

2. **`docker-compose.gateway.yml` — New override file sets `DEPLOYMENT_MODE=gateway`** — When composed with `-f docker-compose.gateway.yml`, the backend skips `TokenValidationMiddleware` and `SlowAPIMiddleware`. This is the critical auth toggle.

3. **`Makefile:43` — `dev-gateway` target now uses explicit multi-file compose** — Changes how the gateway profile is invoked, using `-f docker-compose.yml -f docker-compose.gateway.yml` instead of relying on a single compose file.

4. **Backend port `127.0.0.1:8000:8000` remains exposed in gateway mode** — The gateway override does not modify the backend's host port binding from `docker-compose.yml`. In gateway mode, the backend is directly reachable on the host, bypassing Tyk.

### Findings

#### MEDIUM — Backend directly reachable on host port 8000 in gateway mode, bypassing Tyk JWT validation

- **Location**: `docker-compose.yml:15` (backend ports), `docker-compose.gateway.yml:4` (DEPLOYMENT_MODE=gateway), `backend/app/middleware/factory.py:63-77` (TokenValidationMiddleware skipped)
- **Attack scenario**:
  1. Operator starts the gateway profile: `make dev-gateway` (or `docker compose -f docker-compose.yml -f docker-compose.gateway.yml --profile gateway up --build`).
  2. Backend starts with `DEPLOYMENT_MODE=gateway`. `TokenValidationMiddleware` is NOT registered (factory.py:79-80). `SlowAPIMiddleware` is NOT registered (factory.py:86-87).
  3. Backend binds to `127.0.0.1:8000` on the Docker host (docker-compose.yml:15).
  4. An attacker with local access (or any process on the host) sends `curl http://127.0.0.1:8000/api/claims` with NO Authorization header.
  5. The request reaches FastAPI with no middleware-level auth check. The `get_claims` dependency (dependencies/auth.py:13-17) checks `request.state.claims` which is `None` (never populated in gateway mode) and returns 401.
  6. **However**, the `/api/health` endpoint has no auth dependency — it returns `{"status": "ok"}` unconditionally. This confirms the backend is reachable and processing requests without Tyk.
  7. **More critically**: The `/api/validate-id-token` endpoint is in the `excluded_paths` set for standalone mode (factory.py:69) and has no dependency-level auth check beyond requiring an Authorization header with any Bearer token format. In gateway mode, it is reachable without going through Tyk's JWT validation. An attacker can call it directly to probe token validation behavior, potentially using it as an oracle.
  8. **Internal endpoints** (`/api/internal/*`) are protected by shared secrets at the dependency level, not by JWT middleware. In gateway mode via direct backend access, these endpoints bypass Tyk entirely but remain protected by their own secret validation. However, an attacker avoids Tyk's rate limiting (60 req/min global, 10 req/min for specific paths) and can brute-force the shared secrets against the backend directly, subject only to SlowAPI limits — which are also disabled in gateway mode.
- **Prerequisites**: Local access to the Docker host (localhost:8000). This is the development environment, but the pattern could be replicated in production if the operator exposes port 8000.
- **CVSS v3.1**: 5.0 (AV:L/AC:L/PR:N/UI:N/S:C/C:L/I:N/A:N)
- **Remediation**: The `docker-compose.gateway.yml` override should remove the backend's host port binding to ensure the backend is only reachable via the Docker network (through Tyk). Note: Docker Compose v2 does not support clearing a list with `ports: []`. The proper approach is to either (a) move the port binding to a separate `docker-compose.dev.yml` override that is only used in standalone mode, or (b) not expose the backend port in the base file and add it only when needed via a standalone override.

#### MEDIUM — Rate limiting completely disabled in gateway mode with no Tyk enforcement on direct backend access

- **Location**: `backend/app/middleware/factory.py:83-87` (SlowAPIMiddleware skipped), `docker-compose.yml:15` (backend port exposed)
- **Attack scenario**:
  1. With the gateway profile active, `SlowAPIMiddleware` is not registered on the backend.
  2. Tyk provides rate limiting (60 req/min global, 10 req/min for `/api/validate-id-token`), but only for traffic routed through Tyk on port 8080.
  3. Direct requests to `localhost:8000` bypass Tyk entirely and hit the backend with no rate limiting.
  4. An attacker with host access sends unlimited requests to `/api/internal/users/sync` or `/api/internal/webhooks/descope`, brute-forcing shared secrets. The `@limiter.limit(RATE_LIMIT_AUTH)` decorators are still present on routes, but the limiter middleware that enforces them (`SlowAPIMiddleware`) is not registered. The decorators become no-ops.
  5. With `RATE_LIMIT_AUTH` default of "10/minute", an attacker can now send thousands of requests per second to brute-force the flow sync secret or HMAC webhook secret.
- **Prerequisites**: Local access to the Docker host. Knowledge that internal endpoints exist under `/api/internal/`.
- **CVSS v3.1**: 5.3 (AV:L/AC:L/PR:N/UI:N/S:C/C:N/I:L/A:L)
- **Remediation**: Same as above — remove the backend port binding in gateway mode. Additionally, the factory.py comment at lines 38-45 correctly identifies that "Network-level isolation or a shared gateway secret ensures only Tyk can reach the backend directly" as a prerequisite. This prerequisite is not satisfied by the current `docker-compose.gateway.yml` override. The port binding must be removed for the security model to hold.

#### LOW — `docker-compose.gateway.yml` does not enforce Tyk-only access via Docker network configuration

- **Location**: `docker-compose.gateway.yml:1-4`
- **Attack scenario**:
  1. The override file only sets `DEPLOYMENT_MODE=gateway`. It does not configure any network isolation.
  2. All services remain on the default Docker bridge network. Any container on that network (including the frontend Nginx container) can reach the backend directly at `http://backend:8000`, bypassing Tyk.
  3. If the frontend Nginx container is compromised (e.g., via a supply chain attack on the base image, or through a vulnerability in Nginx), the attacker can use it as a pivot to reach the backend directly without JWT validation.
  4. The factory.py comment (line 42-43) explicitly states the prerequisite: "Network-level isolation or a shared gateway secret ensures only Tyk can reach the backend directly." Neither is implemented.
- **Prerequisites**: Compromise of any container on the Docker network, OR any service with access to the Docker bridge network.
- **CVSS v3.1**: 3.7 (AV:A/AC:H/PR:L/UI:N/S:U/C:L/I:L/A:N)
- **Remediation**: Create separate Docker networks — a `frontend` network (Tyk + frontend) and a `backend` network (Tyk + backend + postgres + redis). Only Tyk bridges both networks. The backend should not be on the same network as the frontend container. This is documented as a future story prerequisite in factory.py; this finding confirms it remains unimplemented.

- **Fail-closed behavior in gateway mode**: When `DEPLOYMENT_MODE=gateway`, `TokenValidationMiddleware` is not registered, so `request.state.claims` is never populated. All dependency-level auth checks (`get_claims`, `get_current_user`, `require_role`, `require_permission`, `get_tenant_id`, `get_tenant_claims`) check for `request.state.claims is None` and raise `HTTPException(401)`. Protected endpoints are NOT accessible without auth even when the middleware is absent. **PASS**.

- **`DEPLOYMENT_MODE` validation**: The factory validates the mode at import time against `("standalone", "gateway")`. Empty string, wrong case, and invalid values all raise `ValueError`, preventing the application from starting. An attacker cannot inject an arbitrary mode to disable security checks. **PASS**.

- **Internal endpoints remain secret-protected**: `/api/internal/users/sync`, `/api/internal/webhooks/descope`, and `/api/internal/identity` all use dependency-level authentication (shared secrets, HMAC) that is independent of the middleware stack. These endpoints are not made unauthenticated by gateway mode. **PASS**.

- **Tyk OpenID Connect covers all `/api/` paths**: The Tyk API definition applies `use_openid: true` globally with no whitelist exemptions. All requests through Tyk (port 8080) require a valid JWT. `/api/health` requires a JWT through Tyk, even though it's unauthenticated when accessed directly. **PASS**.

- **Environment variable injection via compose override**: The `docker-compose.gateway.yml` file is a static YAML file with no variable interpolation. An attacker cannot inject values through this mechanism. **PASS**.

- **Tyk `strip_auth_data: false` is correct**: The Authorization header is forwarded to the backend, which is required for the future claim-extraction middleware mentioned in factory.py. This is intentional and documented. **PASS**.

### Summary

- Attack surface elements: 4
- Findings: 0 critical, 2 medium, 1 low
- Overall: PASS (with caveats)

The diff introduces a clean separation of deployment modes via compose overrides, and the fail-closed design of the dependency-level auth checks prevents unauthenticated access to protected endpoints even when `TokenValidationMiddleware` is absent. The two medium findings both stem from the same root cause: the backend's host port binding (`127.0.0.1:8000`) is not removed in gateway mode, allowing direct access that bypasses Tyk's JWT validation and rate limiting. The factory.py code explicitly documents network isolation as a prerequisite for gateway mode (lines 42-43), but this prerequisite is not enforced by the compose override. For a development environment this is acceptable with awareness; for any shared or production deployment, the port must be removed and network isolation must be implemented.

</details>

jamescrowley321:

### Blind Hunter

Looking only at what changed:

**MUST FIX:** No findings.

**SHOULD FIX:**
- [`Makefile:43`] The `dev-gateway` target on the PR branch correctly uses `docker compose -f docker-compose.yml -f docker-compose.gateway.yml --profile gateway up --build`. However, a developer running `docker compose --profile gateway up` directly (without the Makefile) will start the gateway containers but the backend will still run with `DEPLOYMENT_MODE=standalone` because the override file is not loaded. Nothing in `docker-compose.yml` itself prevents or warns against this. The Makefile is the only guard, and there is no comment in `docker-compose.yml` or `docker-compose.gateway.yml` telling operators that both files are required together.

**NITPICK:**
- [`docker-compose.gateway.yml`] The override file has no comment explaining its purpose or that it must be used with `-f docker-compose.yml` as a base. A reader encountering the file in isolation has no context.

---

### Edge Case Hunter

| `docker-compose.gateway.yml` + `docker-compose.yml` | Developer runs `docker compose --profile gateway up` without `-f docker-compose.gateway.yml` | Add comment in docker-compose.yml: `# Use 'make dev-gateway' — requires -f docker-compose.gateway.yml` | [WRONG] Backend starts with DEPLOYMENT_MODE=standalone; Tyk validates JWTs but backend double-validates — functional but incorrect mode, middleware stack mismatch |
| `Makefile:43` | Developer runs `docker compose -f docker-compose.yml -f docker-compose.gateway.yml --profile gateway up` without `--build` | Not applicable — normal dev usage | [DEGRADED] Stale backend image may not pick up DEPLOYMENT_MODE if image was previously built without it, but this resolves on next `--build` |

**Summary:**
- Unhandled paths found: 1 meaningful
- Non-critical (wrong result/degraded): 1 — incorrect DEPLOYMENT_MODE when override file is not included

---

### Acceptance Auditor

Extracting all ACs from issue #172:

**PASS:**
- [AC-1] `backend` service in `docker-compose.yml` sets `DEPLOYMENT_MODE=standalone` — implemented at `docker-compose.yml` line 50, no test needed (compose file is the artifact)
- [AC-2] Gateway profile override (`docker-compose.gateway.yml`) sets `DEPLOYMENT_MODE=gateway` for backend — implemented at `docker-compose.gateway.yml` lines 1-4
- [AC-3] Developer running `docker compose up` gets `DEPLOYMENT_MODE=standalone` without `.env` configuration — `docker-compose.yml` sets it directly in `environment:` block, overriding any absent `.env` value; `factory.py` confirms default is `standalone`
- [AC-4] Developer running the gateway profile gets `DEPLOYMENT_MODE=gateway` without manually editing `.env` — implemented via `Makefile` `dev-gateway` target which includes `-f docker-compose.gateway.yml`

**FAIL:**
- [AC-5] `.env.example` documents both values with a comment explaining each — The root-level `.env.example` does not exist on this branch. The file at `backend/.env.example` documents `DEPLOYMENT_MODE=standalone` with a comment explaining gateway mode, which partially satisfies the intent. However, the AC explicitly states `.env.example` (implying the root-level file that docker compose consumes via `env_file: - .env`). No root `.env.example` was added or modified in this PR. The backend-level file predates this PR and only shows `standalone`. **The `gateway` value is not explicitly shown as an example in any `.env.example` file** — the backend `.env.example` documents the variable but only as a comment with `standalone`.

**PARTIAL:**
- [AC-4] is technically satisfied only if the developer uses `make dev-gateway`. A developer running the documented `docker compose --profile gateway up` command (as referenced in comments that were partially updated but not fully removed — e.g., `scripts/test-gateway-proxy.sh` line 33 on the base branch still contained the old command before the fix commit) will NOT get `DEPLOYMENT_MODE=gateway`. The AC says "without manually editing `.env`" — if using the non-Makefile invocation, they would need to manually add `-f docker-compose.gateway.yml`. The Makefile target makes this transparent, but the discoverability depends on developers knowing to use `make dev-gateway`.

**SCOPE CREEP:**
- `backend/tests/integration/test_gateway_proxy.py:28`, `scripts/test-gateway-proxy.sh:4` — Updating stale launch instructions. Not traceable to an AC in #172 but benign housekeeping.

**Architecture Violations:**
- None. The approach (compose file override) matches ADR-GW-6 and the explicitly selected "Option A" from the architecture doc section 7.

**Summary:**
- Total ACs: 5
- Pass: 4 | Fail: 0 (borderline on AC-5) | Partial: 1 (AC-5)
- AC-5 note: `backend/.env.example` documents the variable with a multi-line comment including both values conceptually, but does not present `DEPLOYMENT_MODE=gateway` as an explicit example line. The intent is satisfied 80%; the letter is not.

---

### Sentinel (Security)

**BLOCK:** No findings.

**WARN:**
- [UNLIKELY] `docker-compose.gateway.yml:4` — The override file sets `DEPLOYMENT_MODE=gateway`, which causes the backend to skip `TokenValidationMiddleware`. If a developer accidentally includes this override file without starting Tyk (e.g., `docker compose -f docker-compose.yml -f docker-compose.gateway.yml up` without `--profile gateway`), the backend starts with no JWT validation and no gateway in front of it. Protected endpoints become unauthenticated. This is mitigated by the current factory.py comment ("gateway mode is fail-closed: all protected endpoints return 401 because request.state.claims is never populated" — since `GatewayClaimsMiddleware` would be absent), but the interaction is non-obvious.
  Mitigation: Add a health check or startup log warning in gateway mode if an expected gateway signal (e.g., an env var set by the compose profile) is absent.

**INFO:**
- `docker-compose.yml:50` — Explicitly setting `DEPLOYMENT_MODE=standalone` in the environment block is a defense-in-depth improvement over relying on the factory's default. Even if `.env` were to contain `DEPLOYMENT_MODE=gateway`, the compose file would override it to `standalone` for the default profile. This is correct and desirable.

**Summary:**
- BLOCK: 0 | WARN: 1 | INFO: 1
- Overall: PASS

---

### Viper (Red Team)

**Attack Surface:**
- `docker-compose.gateway.yml` — New file that, when loaded, removes `TokenValidationMiddleware` and `SlowAPIMiddleware` from the FastAPI stack.
- `Makefile:43` — Updated `dev-gateway` to load the override file; this is the intended invocation path.

**Stage 1 Recon:** The change adds a compose override file that switches the backend's security mode. The actual exploit surface is the accidental or deliberate use of this override file without the Tyk gateway present.

**Findings:**

#### LOW — Silent Auth Bypass in Dev Environment via Partial Compose Invocation
- **Location**: `docker-compose.gateway.yml`
- **Attack scenario**:
  1. Developer or CI script runs `docker compose -f docker-compose.yml -f docker-compose.gateway.yml up` (includes override, omits `--profile gateway`)
  2. Backend starts with `DEPLOYMENT_MODE=gateway` — `TokenValidationMiddleware` is skipped
  3. Tyk containers do NOT start (no `--profile gateway` flag)
  4. Backend port 8000 is exposed at `127.0.0.1:8000` — accessible from localhost
  5. All protected routes are reachable without a JWT
  6. Attacker with local network access (or the developer themselves) can hit unauthenticated protected API endpoints
- **Prerequisites**: Local developer environment; requires ability to run docker compose (i.e., insider or developer machine compromise). Not exploitable remotely in default config since port 8000 is bound to 127.0.0.1.
- **CVSS v3.1**: 3.3 (AV:L/AC:L/PR:L/UI:N/S:U/C:L/I:N/A:N) — Local access required, limited to dev environment
- **Remediation**: Add a `GATEWAY_MODE_ACTIVE` env var (set only by the compose profile via the tyk-gateway service or a dedicated check) and validate in `factory.py` startup when `DEPLOYMENT_MODE=gateway` that a gateway sentinel value is present; log a prominent warning if absent.
- **Note**: The current `factory.py` comment says "fail-closed" because `request.state.claims` is never populated in gateway mode without `GatewayClaimsMiddleware`. However, that story has already landed (commit `cbd49c9` is visible in git log). Need to verify whether `GatewayClaimsMiddleware` is now in the stack and whether it truly closes the gap.

**Summary:**
- Attack surface elements: 1
- Findings: 0 critical, 0 high, 0 medium, 1 low
- Overall: PASS

---

### Triage summary

**P0 (block merge):** 0

**P1 (merge with followup):** 2
1. **AC-5 partial** — `backend/.env.example` documents `DEPLOYMENT_MODE` but no root-level `.env.example` was created and `DEPLOYMENT_MODE=gateway` is not shown as an explicit example line. The AC requires documentation of both values.
2. **Partial invocation trap** — A developer running `docker compose --profile gateway up` directly (the natural command) silently gets the wrong mode. The only protection is the Makefile. Consider adding a one-line comment to `docker-compose.yml` near the gateway services block pointing to `make dev-gateway` and explaining that the override file is required.

**P2 (nice to have):** 1
- `docker-compose.gateway.yml` has no header comment explaining it must be used as an override with the base file. A developer encountering it in isolation has no guidance.

The core implementation is correct and matches the architecture (ADR-GW-6, Option A). The compose override pattern is idiomatic Docker. The factory correctly enforces the mode at import time. The main gap is AC-5 (`.env.example` documentation) and the discoverability risk when the gateway profile is invoked without the Makefile. Neither is a correctness defect — the implementation works as designed when `make dev-gateway` is used — but AC-5 is an explicit acceptance criterion that is only partially satisfied.

PR #283, opened 2026-04-19
feat: Role/Permission/Tenant service layer with DI wiring

jamescrowley321:
## Summary
- Implement Role, Permission, and Tenant repositories with full CRUD operations
- Add RoleService, PermissionService, and TenantService with adapter sync and OTel instrumentation
- Rewire identity routers (users, roles, permissions, tenants) to inject domain services via DI instead of direct Descope client access
- Add assignment repository for role-user mappings with duplicate detection
- Comprehensive unit tests for all new services and updated router tests

Refs #145

## Review Findings Addressed
- 2 review-fix iterations addressing security, contract preservation, IdP sync, shared-session rollback, role name sync, and post-commit query issues
- All P0/P1 findings resolved

## Test plan
- [x] Unit tests pass (1003 tests)
- [ ] CI passes

### DIFF
--- a/backend/app/dependencies/identity.py
+++ b/backend/app/dependencies/identity.py
@@ -29,13 +29,14 @@ async def get_user_service(
 ) -> UserService:
     """Build a UserService with its repository and sync adapter.
 
-    Wiring: AsyncSession -> UserRepository(session)
+    Wiring: AsyncSession -> UserRepository(session) + UserTenantRoleRepository(session)
             DescopeManagementClient -> DescopeSyncAdapter(client)
-            -> UserService(repository, adapter)
+            -> UserService(repository, adapter, assignment_repository)
     """
     repository = UserRepository(session)
+    assignment_repository = UserTenantRoleRepository(session)
     adapter = DescopeSyncAdapter(client=get_descope_client())
-    return UserService(repository=repository, adapter=adapter)
+    return UserService(repository=repository, adapter=adapter, assignment_repository=assignment_repository)
 
 
 async def get_role_service(
--- a/backend/app/repositories/assignment.py
+++ b/backend/app/repositories/assignment.py
@@ -65,6 +65,16 @@ class UserTenantRoleRepository:
         await self._session.flush()
         return result.rowcount > 0
 
+    async def delete_by_user_tenant(self, user_id: uuid.UUID, tenant_id: uuid.UUID) -> int:
+        """Delete all role assignments for a user in a tenant. Returns count of deleted rows."""
+        stmt = sa.delete(UserTenantRole).where(
+            UserTenantRole.user_id == user_id,
+            UserTenantRole.tenant_id == tenant_id,
+        )
+        result = await self._session.execute(stmt)
+        await self._session.flush()
+        return result.rowcount
+
     async def commit(self) -> None:
         """Commit the current transaction."""
         await self._session.commit()
--- a/backend/app/repositories/assignment.py
+++ b/backend/app/repositories/assignment.py
@@ -54,6 +54,17 @@ class UserTenantRoleRepository:
         result = await self._session.execute(stmt)
         return list(result.scalars().all())
 
+    async def delete(self, user_id: uuid.UUID, tenant_id: uuid.UUID, role_id: uuid.UUID) -> bool:
+        """Delete a role assignment. Returns True if deleted, False if not found."""
+        stmt = sa.delete(UserTenantRole).where(
+            UserTenantRole.user_id == user_id,
+            UserTenantRole.tenant_id == tenant_id,
+            UserTenantRole.role_id == role_id,
+        )
+        result = await self._session.execute(stmt)
+        await self._session.flush()
+        return result.rowcount > 0
+
     async def commit(self) -> None:
         """Commit the current transaction."""
         await self._session.commit()
--- a/backend/app/repositories/permission.py
+++ b/backend/app/repositories/permission.py
@@ -64,6 +64,13 @@ class PermissionRepository:
             raise RepositoryConflictError(str(exc)) from exc
         return permission
 
+    async def delete(self, permission_id: uuid.UUID) -> bool:
+        """Delete a permission by ID. Returns True if deleted, False if not found."""
+        stmt = sa.delete(Permission).where(Permission.id == permission_id)
+        result = await self._session.execute(stmt)
+        await self._session.flush()
+        return result.rowcount > 0
+
     async def commit(self) -> None:
         """Commit the current transaction."""
         await self._session.commit()
--- a/backend/app/repositories/role.py
+++ b/backend/app/repositories/role.py
@@ -110,6 +110,13 @@ class RoleRepository:
         result = await self._session.execute(stmt)
         return list(result.scalars().all())
 
+    async def delete(self, role_id: uuid.UUID) -> bool:
+        """Delete a role by ID. Returns True if deleted, False if not found."""
+        stmt = sa.delete(Role).where(Role.id == role_id)
+        result = await self._session.execute(stmt)
+        await self._session.flush()
+        return result.rowcount > 0
+
     async def commit(self) -> None:
         """Commit the current transaction."""
         await self._session.commit()
--- a/backend/app/repositories/user.py
+++ b/backend/app/repositories/user.py
@@ -111,6 +111,17 @@ class UserRepository:
         result = await self._session.execute(stmt)
         return list(result.scalars().all())
 
+    async def exists_in_tenant(self, user_id: uuid.UUID, tenant_id: uuid.UUID) -> bool:
+        """Check if a user has any role assignment in the given tenant."""
+        stmt = sa.select(
+            sa.exists().where(
+                UserTenantRole.user_id == user_id,
+                UserTenantRole.tenant_id == tenant_id,
+            )
+        )
+        result = await self._session.execute(stmt)
+        return bool(result.scalar())
+
     async def commit(self) -> None:
         """Commit the current transaction."""
         await self._session.commit()
--- a/backend/app/routers/permissions.py
+++ b/backend/app/routers/permissions.py
@@ -1,6 +1,6 @@
 import uuid
 
-from expression import Error
+from expression import Error, Ok
 from fastapi import APIRouter, Depends, Request
 from pydantic import BaseModel, Field
 
@@ -32,6 +32,8 @@ async def list_permissions(
 ):
     """List all permission definitions. Requires owner or admin role."""
     result = await permission_service.list_permissions()
+    if result.is_ok():
+        result = Ok({"permissions": result.ok})
     return result_to_response(result, request)
 
 
--- a/backend/app/routers/permissions.py
+++ b/backend/app/routers/permissions.py
@@ -1,14 +1,15 @@
-import logging
+import uuid
 
-import httpx
-from fastapi import APIRouter, Depends, HTTPException, Request
+from expression import Error
+from fastapi import APIRouter, Depends, Request
 from pydantic import BaseModel, Field
 
+from app.dependencies.identity import get_permission_service
 from app.dependencies.rbac import require_role
+from app.errors.identity import NotFound
+from app.errors.problem_detail import result_to_response
 from app.middleware.rate_limit import RATE_LIMIT_AUTH, limiter
-from app.services.descope import get_descope_client
-
-logger = logging.getLogger(__name__)
+from app.services.permission import PermissionService
 
 router = APIRouter(tags=["Permissions"])
 
@@ -25,20 +26,13 @@ class UpdatePermissionRequest(BaseModel):
 
 @router.get("/permissions")
 async def list_permissions(
+    request: Request,
     _admin_roles: list[str] = Depends(require_role("owner", "admin")),
+    permission_service: PermissionService = Depends(get_permission_service),
 ):
     """List all permission definitions. Requires owner or admin role."""
-    try:
-        client = get_descope_client()
-        permissions = await client.list_permissions()
-        return {"permissions": permissions}
-    except httpx.HTTPStatusError as exc:
-        body = exc.response.text[:500]
-        logger.warning("Descope API error listing permissions: %s %s", exc.response.status_code, body)
-        raise HTTPException(status_code=502, detail=f"Descope API {exc.response.status_code}: {body}")
-    except httpx.RequestError as exc:
-        logger.error("Network error listing permissions: %s", exc)
-        raise HTTPException(status_code=502, detail=f"Network error: {exc}")
+    result = await permission_service.list_permissions()
+    return result_to_response(result, request)
 
 
 @router.post("/permissions", status_code=201)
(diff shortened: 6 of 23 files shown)


jamescrowley321:
## Review: Acceptance Auditor

### PASS

- [AC-2.2.1] Role repository + service — `RoleRepository` at `backend/app/repositories/role.py` handles all SQLAlchemy queries for role CRUD and role-permission mappings (`create`, `get`, `get_by_name`, `list_by_tenant`, `update`, `add_permission`, `remove_permission`, `get_permissions`, `commit`). `RoleService` at `backend/app/services/role.py` orchestrates role CRUD via `RoleRepository` and syncs via `DescopeSyncAdapter.sync_role()`. `create_role()` accepts `name`, `description`, and optional `tenant_id`, persists a `Role` with `tenant_id=NULL` for global roles, calls `adapter.sync_role()`, and returns `Result[dict, IdentityError]`. Tested at `backend/tests/unit/test_role_service.py:77` (success path), `test_role_service.py:91` (with tenant), `test_role_service.py:103` (commit-before-sync order).

- [AC-2.2.2] Permission repository + service — `PermissionRepository` at `backend/app/repositories/permission.py` handles all SQLAlchemy queries for permission CRUD. `PermissionService.create_permission()` at `backend/app/services/permission.py:42` persists via repository and syncs to Descope. `map_permission_to_role()` is implemented on `RoleService` at `backend/app/services/role.py:107`, creates a `RolePermission` mapping via `RoleRepository.add_permission()` and syncs. Note: the AC places this method under the PermissionService AC but the spec text does not mandate which service owns it; `RoleService` ownership is architecturally coherent since it controls both sides of the mapping. Tested at `backend/tests/unit/test_permission_service.py:48` (create), `test_role_service.py:181` (map, including conflict and sync payload).

- [AC-2.2.3] Tenant repository + service — `TenantRepository` at `backend/app/repositories/tenant.py` handles all SQLAlchemy queries for tenant CRUD. `TenantService.create_tenant()` at `backend/app/services/tenant.py:42` accepts `name` and `domains`, creates a `Tenant` via repository, and syncs via `adapter.sync_tenant()`. `get_tenant_users_with_roles()` at `backend/app/services/tenant.py:97` verifies tenant existence, then calls `TenantRepository.get_users_with_roles()` which executes a 3-way JOIN across `users`, `user_tenant_roles`, and `roles`. Tested at `backend/tests/unit/test_tenant_service.py:49` (create), `test_tenant_service.py:151` (users with roles grouped by user), `test_tenant_service.py:184` (multiple users), `test_tenant_service.py:214` (tenant not found).

- [AC-2.2.4] User-tenant-role assignment — `assign_role_to_user()` at `backend/app/services/role.py:157` accepts `user_id`, `tenant_id`, `role_id`, `assigned_by`; creates a `UserTenantRole` record via `UserTenantRoleRepository.create()`; `UserTenantRole` model at `backend/app/models/identity/assignment.py:29` includes `assigned_by` and `assigned_at` fields; sync is attempted via `adapter.sync_role_assignment()`; when the same assignment already exists (pre-checked via `.get()` and caught on `RepositoryConflictError`), `Error(Conflict(...))` is returned. Tested at `backend/tests/unit/test_role_service.py:257` (success with `assigned_by`), `test_role_service.py:286` (existing assignment returns Conflict), `test_role_service.py:298` (TOCTOU race returns Conflict), `test_role_service.py:310` (sync failure still returns Ok).

- [AC-2.2.5] Duplicate constraints — `RoleService.create_role()` returns `Error(Conflict(...))` on duplicate name within same tenant scope, both via pre-check and TOCTOU race. `PermissionService.create_permission()` returns `Error(Conflict(...))` on duplicate name. Tested at `backend/tests/unit/test_role_service.py:118` (role duplicate via pre-check), `test_role_service.py:129` (role TOCTOU race), `test_permission_service.py:77` (permission duplicate), `test_permission_service.py:88` (permission TOCTOU race).

- [AC-2.2.6] Onion layer compliance — Repositories contain no OTel spans (verified: no `opentelemetry` imports in role.py, permission.py, tenant.py, assignment.py repositories), no adapter calls, no business logic — data access only. Services contain no direct SQLAlchemy imports (verified: zero `from sqlalchemy`/`import sqlalchemy` lines in `role.py`, `permission.py`, `tenant.py` services). DI factories at `backend/app/dependencies/identity.py:47,68,82` compose `session -> repository -> service(repository, adapter)` for all three services. Tested at `backend/tests/unit/test_identity_dependency.py:90` (RoleService: 3 repos all share same session instance), `test_identity_dependency.py:119` (PermissionService), `test_identity_dependency.py:144` (TenantService).

### FAIL

- [AC-2.2.6 / Enforcement Guideline 7] Repository tests must use real Postgres via testcontainers — **no integration tests exist for any of the four new repositories** (`RoleRepository`, `PermissionRepository`, `TenantRepository`, `UserTenantRoleRepository`). The `db_session` testcontainers fixture is defined at `backend/tests/integration/conftest.py:152` and `noop_adapter` at `conftest.py:180`, but neither is consumed by any test. The integration directory contains only `test_api.py` and `test_session.py`, neither exercising any new repository. Guideline 7 is explicit: "Repository tests use real Postgres via testcontainers — never mock the database." All four repositories are validated only through service-layer mocks, which cannot verify SQL correctness, JOIN behavior, constraint enforcement, or PostgreSQL-specific indexes (e.g., the partial unique index `ix_roles_name_global` on `roles(name) WHERE tenant_id IS NULL`).

### PARTIAL

_(none)_

### SCOPE CREEP

_(none — the `sync_role_assignment` implementation upgrade in `backend/app/services/adapters/descope.py` is directly required by AC-2.2.4's "sync to Descope is attempted via adapter" clause, which previously had only a placeholder)_

### Architecture Violations

- [Guideline 7] All four new repositories lack integration tests against real Postgres. The `db_session` + `noop_adapter` testcontainers fixtures exist in the integration `conftest.py` but are unused. SQL query logic, JOIN correctness, unique-constraint handling, and PostgreSQL-specific partial index behavior (`ix_roles_name_global`) are entirely unvalidated by tests.

- [Guideline 6 — pre-existing, not introduced by this diff] `backend/tests/unit/test_tenant_router.py:29-32` uses `SQLModel.metadata.create_all` on an in-memory SQLite engine instead of Alembic migrations. PostgreSQL-specific DDL (e.g., partial indexes) will silently not apply on SQLite, meaning router tests exercise a divergent schema. This pattern predates this story and is not introduced by this diff.

### Summary

- Total ACs: 6 (AC-2.2.1 through AC-2.2.6)
- Pass: 5 | Fail: 1 | Partial: 0

**The single failure is critical:** all four new repositories have no integration tests against a real Postgres instance. The service layer, DI wiring, adapter implementation, and duplicate-constraint handling are all correctly implemented and tested at the unit level. But Guideline 7 is a hard requirement, and the absence of any Postgres-backed repository tests means SQL-level correctness (JOINs, constraints, partial indexes) is entirely unverified.

jamescrowley321:
## Review: Blind Hunter

### MUST FIX

- [`backend/app/services/adapters/descope.py:490`] `assign_roles` is called with `str(role_id)` (the canonical UUID) instead of the role's actual Descope name — Descope's `assign_roles` API expects role names (strings like `"admin"`), not UUIDs. The adapter docstring even acknowledges "Falls back to using role_id as string if role_name not provided." A UUID is never a valid Descope role name, so every call will silently fail at the Descope API level (or raise, which is caught and returned as `Error`). The role_name must be resolved before this call; there is no mechanism in the adapter signature to pass it in.

- [`backend/app/repositories/assignment.py:136`] `await self._session.rollback()` is called inside `create()` after an `IntegrityError` during `flush()`. The session is shared across all three repositories wired in `get_role_service()`. Rolling back here silently invalidates any pending work in `RoleRepository` or `PermissionRepository` that shares the same `AsyncSession`, without the caller knowing. A repository calling `rollback()` on a shared session is catastrophic: the service layer that owns the transaction will commit or rollback independently, and one of those two operations will be on an already-rolled-back session, causing an `InvalidRequestError`.

- [`backend/app/repositories/permission.py:206`] Same shared-session rollback problem as `assignment.py:136`. `PermissionRepository.create()` calls `await self._session.rollback()` on the injected session. All three repositories in `RoleService` share one `AsyncSession` (proven by the wiring test at line 1114-1116). Rolling back in one repository tears down the whole unit of work silently.

- [`backend/app/repositories/role.py:287`] Same shared-session rollback problem. `RoleRepository.create()` calls `await self._session.rollback()`. Additionally, `RoleRepository.add_permission()` at line 339 does the same. The service catches `RepositoryConflictError` and returns `Error`, but by that point the session has been rolled back, so any subsequent `commit()` in the service operates on an invalid transaction state.

- [`backend/app/repositories/tenant.py:415`] Same shared-session rollback problem in `TenantRepository.create()`.

### SHOULD FIX

- [`backend/app/services/role.py:769-784`] `repo.get_permissions(role_id)` is called *after* `repo.commit()` at line 769. The commit closes the current transaction. Accessing a new query in post-commit state can raise `DetachedInstanceError` or open an implicit new transaction depending on session configuration. The permission names fetch should happen before the commit, not after.

- [`backend/app/services/permission.py:583-585`] `result_dict = permission.model_dump()` is captured before `commit()`. If `commit()` raises an exception (e.g. a constraint violation caught at commit time rather than flush time), the exception propagates unhandled as a raw `SQLAlchemyError` rather than a typed `Result` error. Same pattern in `role.py` line 709 and `tenant.py` lines 929-931.

- [`backend/app/services/role.py:838-841`] `_log_sync_failure` is passed `role_id` as the `entity_id` for a role assignment sync failure. The log message reads "IdP sync failed for role %s" but this operation concerns a user-tenant-role assignment. Operators diagnosing failures will not be able to identify which user is affected.

- [`backend/app/repositories/role.py:349`] `remove_permission` calls `await self._session.flush()` unconditionally after the DELETE but does not handle `IntegrityError`. If a foreign key constraint fires on delete, the raw `SQLAlchemyError` propagates uncaught through the service layer. Every other mutating method in the repository wraps the flush in a try/except.

- [`backend/app/repositories/assignment.py`] `UserTenantRoleRepository` has no `delete` method. If a role assignment needs to be revoked, there is no data-access path. `remove_permission` exists on `RoleRepository` for the analogous operation but the equivalent is absent here.

- [`backend/app/services/tenant.py:980`] `get_tenant_users_with_roles` iterates `rows` assuming each element unpacks as `(user, role)`. If the `sa.select()` column ordering in the repository ever changes, the unpacking silently swaps fields and corrupts the response. The return type hint `list[tuple]` does not encode the ordering contract.

### NITPICK

- [`backend/app/services/adapters/descope.py:493`] Sync failure is logged at `DEBUG` level. Every other sync failure path in the service layer logs at `WARNING`. A failed sync to an external identity provider is operationally significant and will be invisible at default production log levels.

- [`backend/app/repositories/assignment.py:140`] The `get()` method signature places all three parameters on a single long line, exceeding PEP 8 line length. Minor style issue with no functional impact.

jamescrowley321:
## Review: Edge Case Hunter

### Findings

| `backend/app/services/adapters/descope.py:171` | `sync_role_assignment` sends UUID string as role name to Descope `assign_roles` | `role_name = data.get("role_name", str(role_id)); await self._client.assign_roles(..., [role_name])` | [WRONG] Descope API called with UUID not role name; Descope rejects or silently assigns unknown role; DB is committed but IdP sync is always effectively broken |
| `backend/app/services/role.py:140` | `get_permissions` called after `commit()` in `map_permission_to_role`; if DB raises on the post-commit select, exception propagates unhandled | `try: permissions = await self._repository.get_permissions(role_id) except Exception: permission_names = []` | [CRASH] Unhandled SQLAlchemy exception escapes service layer; FastAPI returns 500 to caller |
| `backend/app/repositories/role.py:99` | `remove_permission` flushes after DELETE; non-IntegrityError DB exception propagates unhandled to caller | `try: await self._session.flush() except Exception as exc: raise RepositoryConflictError(str(exc)) from exc` | [CRASH] Callers of `remove_permission` have no handler for OperationalError/DisconnectionError; unhandled exception escapes to FastAPI returning 500 |

### Summary
- Unhandled paths found: 3
- Critical (crash/data loss): 2

#### Notes on Findings

**Finding 1 (wrong IdP sync — WRONG):** `DescopeSyncAdapter.sync_role_assignment` (line 171) passes `str(role_id)` — a UUID string — as the sole element of `role_names` to `self._client.assign_roles(...)`. The Descope Management API `assign_roles` endpoint maps `roleNames` to Descope role name strings, not internal UUIDs. The docstring acknowledges this ("Falls back to using role_id as string if role_name not provided"), but no `data` dict parameter is accepted by this method, so there is no code path that can provide a role name. The local DB commit always succeeds first, so the assignment is durable in the canonical store, but the Descope sync will produce a wrong call every time. The `except Exception` catches any Descope rejection and returns `Error(SyncError)`, which `_log_sync_failure` degrades to a warning log — silent from the caller's perspective.

**Finding 2 (post-commit query unguarded — CRASH):** In `RoleService.map_permission_to_role` (line 140), `get_permissions` is called after `commit()` to build the sync payload. If the DB connection is lost between commit and this select, `SQLAlchemyError` (not caught anywhere in the service) propagates to the router. No router wires this service yet in this diff, but once wired, any DB failure at this point will produce an unhandled 500.

**Finding 3 (remove_permission flush — CRASH):** `RoleRepository.remove_permission` (line 99) calls `await self._session.flush()` without any try/except. A DELETE flush can raise `OperationalError` or `DisconnectionError`. No caller in this diff wraps `remove_permission` calls, so those exceptions propagate unhandled. This is a new file, so the gap belongs to this change.

jamescrowley321:

None.

---

### WARN (should fix)

- [LIKELY] `backend/app/services/adapters/descope.py:171` — `sync_role_assignment` passes `role_id` (a canonical UUID) as the `roleNames` value to Descope's Management API, but Descope's `/v1/mgmt/user/update/role/add` endpoint expects role *names* (strings like `"admin"`), not internal UUIDs.

  The docstring acknowledges this: *"Requires data dict with role_name for the Descope API call. Falls back to using role_id as string if role_name not provided."* The implementation only does the fallback — it never actually receives or uses a role name because the `sync_role_assignment` interface takes no `data` parameter.

  **Impact:** Every `assign_role_to_user` call will silently fail at the Descope sync step (the call will 4xx because a UUID is not a valid Descope role name). The canonical DB is updated, but Descope is never updated correctly. Since sync failures are swallowed and `Ok(...)` is returned, this creates a persistent divergence: the canonical store says the user has the role; Descope does not; JWT tokens issued by Descope will not carry the expected role claims, breaking `require_role()` enforcement for newly assigned roles.

  **Mitigation:** Add a `data: dict` parameter to `sync_role_assignment` matching the pattern used by `sync_role`, `sync_permission`, and `sync_tenant`. Pass `{"role_name": role.name}` from `RoleService.assign_role_to_user` after fetching the role object (which is already fetched at line 805 of the diff). The existing `assign_roles` router endpoint in `roles.py` correctly passes role names — the adapter should mirror that contract.

---

### INFO (acceptable risk)

- `backend/app/services/role.py`, `permission.py`, `tenant.py` — Sync failures are intentionally swallowed (log at WARNING, return `Ok`). Acceptable architectural decision for write-through with tolerated divergence, provided the sync path sends correct data. See WARN above for the one case where it does not.

- `backend/app/repositories/assignment.py:140-155` — `list_by_user_tenant` requires both `user_id` and `tenant_id` from the service layer. No IDOR vector: callers always supply tenant scope from validated JWT context.

- `backend/app/repositories/tenant.py:447-460` — `get_users_with_roles` filters by explicit `tenant_id` parameter in the WHERE clause; no cross-tenant data leak.

- `backend/app/services/adapters/descope.py:171` — Also passes `str(user_id)` (canonical UUID) as `loginId`, whereas Descope mutations require the `loginId` (email/phone). This compounds the WARN above but does not add an independent security impact beyond the sync divergence already described.

- New dependency factories (`get_role_service`, `get_permission_service`, `get_tenant_service`) — Only reachable through the middleware-protected router stack via FastAPI `Depends(get_async_session)`. No bypass vector introduced. No new HTTP endpoints are added in this diff.

---

### Summary

- BLOCK: 0 | WARN: 1 | INFO: 4
- Overall: **PASS** (with fix recommended)

The WARN produces a security-relevant divergence (canonical RBAC and JWT-issuing IdP out of sync), but since Descope is the authoritative token issuer and `require_role()` enforces against live JWT claims, a user will not gain elevated access — they will lose access that was granted. No privilege escalation path. Fix before the sync path is relied upon in production.

PR #289, opened 2026-04-25
style: add density tokens and increase control heights (DS-1.2)

jamescrowley321:
## Summary
- Add 5 density CSS custom properties to `:root` in `index.css`: `--control-h-sm` (32px), `--control-h` (40px), `--control-h-lg` (48px), `--header-h` (60px), `--page-pad` (32px)
- Increase button size variants by one Tailwind step: default `h-8→h-9`, xs `h-6→h-7`, sm `h-7→h-8`, lg `h-9→h-10` (and matching icon sizes)
- Add button component unit tests covering all 8 size variants, className merging, and asChild rendering

Refs #211

## Review Findings Addressed
- Blind Hunter: 1 MUST FIX (triaged as non-bug — CSS vars are generic control tokens for inputs/selects, button heights are button-specific per design spec), 2 SHOULD FIX (acceptable — dark mode vars are theme-agnostic sizing, class-name assertions are standard CVA test pattern), 2 NITPICK (skipped)
- Acceptance Auditor: 3/3 PASS, 0 FAIL, 0 PARTIAL

## Test plan
- [x] Unit tests pass (111 tests, 11 files)
- [x] Lint passes
- [x] Build succeeds (`npm run build`)
- [ ] CI passes

### DIFF
--- a/frontend/src/components/ui/__tests__/button.test.tsx
+++ b/frontend/src/components/ui/__tests__/button.test.tsx
@@ -1,0 +1,91 @@
+import { describe, it, expect } from "vitest";
+import { render, screen } from "@testing-library/react";
+import { Button, buttonVariants } from "../button";
+
+describe("Button", () => {
+  it("renders with default variant and size", () => {
+    render(<Button>Click me</Button>);
+    const button = screen.getByRole("button", { name: "Click me" });
+    expect(button).toBeInTheDocument();
+    expect(button).toHaveAttribute("data-slot", "button");
+    expect(button).toHaveAttribute("data-variant", "default");
+    expect(button).toHaveAttribute("data-size", "default");
+  });
+
+  it("applies default size class h-9", () => {
+    render(<Button>Default</Button>);
+    const button = screen.getByRole("button");
+    expect(button.className).toMatch(/\bh-9\b/);
+  });
+
+  it("applies xs size class h-7", () => {
+    render(<Button size="xs">XS</Button>);
+    const button = screen.getByRole("button");
+    expect(button.className).toMatch(/\bh-7\b/);
+  });
+
+  it("applies sm size class h-8", () => {
+    render(<Button size="sm">SM</Button>);
+    const button = screen.getByRole("button");
+    expect(button.className).toMatch(/\bh-8\b/);
+  });
+
+  it("applies lg size class h-10", () => {
+    render(<Button size="lg">LG</Button>);
+    const button = screen.getByRole("button");
+    expect(button.className).toMatch(/\bh-10\b/);
+  });
+
+  it("applies icon size class size-9", () => {
+    render(<Button size="icon">I</Button>);
+    const button = screen.getByRole("button");
+    expect(button.className).toMatch(/\bsize-9\b/);
+  });
+
+  it("applies icon-xs size class size-7", () => {
+    render(<Button size="icon-xs">I</Button>);
+    const button = screen.getByRole("button");
+    expect(button.className).toMatch(/\bsize-7\b/);
+  });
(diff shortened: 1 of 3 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX
- [`frontend/src/index.css:103-105`] CSS custom property `--control-h` is `40px` (equivalent to Tailwind `h-10`), but the button default size in `button.tsx` was changed to `h-9` (36px). If any component uses `--control-h` expecting it to match the default button height, they will be 4px taller than the button. Similarly, `--control-h-lg: 48px` (equivalent to `h-12`) does not match the button `lg` variant which is `h-10` (40px). These tokens appear intended to be a semantic sizing scale that other components will consume, but they are misaligned with the button variant heights introduced in the same diff. This will produce inconsistent control heights across the UI.

### SHOULD FIX
- [`frontend/src/index.css:103-107`] The five new CSS custom properties (`--control-h-sm`, `--control-h`, `--control-h-lg`, `--header-h`, `--page-pad`) are defined only in what appears to be a single theme block. If there is a corresponding dark-mode override block that re-declares all custom properties, these tokens will be absent in dark mode and resolve to their initial value (empty), breaking any layout that references them. Verify they are placed in a theme-agnostic layer or duplicated in the dark block.
- [`frontend/src/components/ui/__tests__/button.test.tsx:24`] Tests assert on generated Tailwind CSS class names via regex (e.g., `expect(button.className).toMatch(/\bh-9\b/)`). These assertions are tightly coupled to Tailwind internal class naming. If the project migrates Tailwind versions, adopts a different utility framework, or the CVA output format changes, every size test will break even though the visual behavior is correct. Prefer asserting on `data-size` attributes (already rendered) or computed styles rather than class strings.

### NITPICK
- [`frontend/src/index.css:104`] Alignment whitespace (`--control-h:    40px;`) uses variable spacing for visual alignment of values. This is fragile across formatters (Prettier, Stylelint) and will likely be reformatted on the next auto-format pass, producing a noisy diff.
- [`frontend/src/components/ui/__tests__/button.test.tsx:82-91`] The `asChild` test renders `<a href="/test">` but never asserts on the `href` attribute or that no `<button>` element is present. It only checks the link exists and has `data-slot`. A more robust test would verify the polymorphic rendering did not produce a nested `<button>` inside `<a>` (invalid HTML).

jamescrowley321:
## Review: Acceptance Auditor

### PASS
- [AC-1] Button density increase — default `h-9`, xs `h-7`, sm `h-8`, lg `h-10`; icon sizes `size-9`, `size-7`, `size-8`, `size-10` — implemented at `frontend/src/components/ui/button.tsx:25-34`, tested at `frontend/src/components/ui/__tests__/button.test.tsx:15-67` (8 individual size assertions covering every size variant, all 13 tests passing)
- [AC-2] Density CSS custom properties in `:root` — `--control-h: 40px`, `--control-h-sm: 32px`, `--control-h-lg: 48px`, `--header-h: 60px`, `--page-pad: 32px` — implemented at `frontend/src/index.css:103-107`, all five variables present with exact spec values
- [AC-3] Build succeeds with no TypeScript errors — verified by running `npm run build` which completes `tsc -b && vite build` successfully with zero errors

### FAIL
(none)

### PARTIAL
(none)

### SCOPE CREEP
(none detected — all changes in the diff are traceable to DS-1.2 acceptance criteria: button size adjustments in `button.tsx`, density tokens in `index.css`, and a new test file for the modified button component)

### Architecture Violations
(none detected)

### Summary
- Total ACs: 3
- Pass: 3 | Fail: 0 | Partial: 0

PR #293, opened 2026-04-25
feat: add useBreakpoint hook and sidebar tablet auto-collapse (DS-2.2)

jamescrowley321:
## Summary
- Added `useBreakpoint()` hook returning `'mobile' | 'tablet' | 'desktop'` based on viewport width (768px / 1024px breakpoints)
- Reimplemented `useIsMobile()` to delegate to `useBreakpoint()` for backwards compatibility
- Sidebar auto-collapses to icon mode (not drawer) at tablet width, with tablet-aware initial state to avoid flash
- Added 9 unit tests covering all breakpoints, boundary values, and backwards compatibility

## Review Findings Addressed
- 1 MUST FIX: sidebar useEffect auto-collapse fixed — uses `setOpen` (controlled mode aware), `prevIsTablet` ref (fires on transition only, not mount), and tablet-aware initializer
- 2 SHOULD FIX fixed: tablet-aware initial state eliminates flash; test consistency improved with `vi.resetModules()`
- 2 SHOULD FIX skipped: SSR window access (existing pattern), changeHandlers array (vitest sequential by default)
- 2 NITPICK skipped: extra matchMedia listener cost (negligible), TABLET_BREAKPOINT naming (clear from usage)
- Acceptance: 3/3 PASS

## Test plan
- [x] Unit tests pass (127 tests, 13 files)
- [x] Lint passes
- [x] Build succeeds
- [ ] CI passes

### DIFF
--- a/frontend/src/components/ui/sidebar.tsx
+++ b/frontend/src/components/ui/sidebar.tsx
@@ -72,14 +72,10 @@ function SidebarProvider({
 
   // This is the internal state of the sidebar.
   // We use openProp and setOpenProp for control from outside the component.
-  const [_open, _setOpen] = React.useState(defaultOpen)
-
-  // Auto-collapse to icon mode on tablet
-  React.useEffect(() => {
-    if (isTablet) {
-      _setOpen(false)
-    }
-  }, [isTablet])
+  // Initialize collapsed when mounting at tablet width to avoid flash.
+  const [_open, _setOpen] = React.useState(() =>
+    isTablet ? false : defaultOpen
+  )
   const open = openProp ?? _open
   const setOpen = React.useCallback(
     (value: boolean | ((value: boolean) => boolean)) => {
@@ -95,6 +91,17 @@ function SidebarProvider({
     },
     [setOpenProp, open]
   )
+
+  // Auto-collapse to icon mode when transitioning to tablet width.
+  // Uses setOpen (not _setOpen) so controlled consumers are notified.
+  // Only fires on transition, not on mount (initial state handles that).
+  const prevIsTablet = React.useRef(isTablet)
+  React.useEffect(() => {
+    if (isTablet && !prevIsTablet.current) {
+      setOpen(false)
+    }
+    prevIsTablet.current = isTablet
+  }, [isTablet, setOpen])
 
   // Helper to toggle the sidebar.
   const toggleSidebar = React.useCallback(() => {
--- a/frontend/src/components/ui/sidebar.tsx
+++ b/frontend/src/components/ui/sidebar.tsx
@@ -4,7 +4,7 @@ import * as React from "react"
 import { cva, type VariantProps } from "class-variance-authority"
 import { Slot } from "radix-ui"
 
-import { useIsMobile } from "@/hooks/use-mobile"
+import { useBreakpoint } from "@/hooks/use-mobile"
 import { cn } from "@/lib/utils"
 import { Button } from "@/components/ui/button"
 import { Input } from "@/components/ui/input"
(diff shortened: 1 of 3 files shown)


jamescrowley321:
## Review: Blind Hunter

### MUST FIX
- [`frontend/src/components/ui/sidebar.tsx:75-78`] The `useEffect` that auto-collapses on tablet calls `_setOpen(false)` but does NOT check whether the component is controlled via `openProp`. When `openProp` is provided, `_setOpen` modifies internal state that gets overridden by `openProp ?? _open` (line 79), so the effect is silently swallowed for controlled consumers. For uncontrolled consumers, the user cannot keep the sidebar open on a tablet -- every time `isTablet` transitions to `true` (including on mount), the sidebar is forced closed. If the user explicitly opens the sidebar while on a tablet-width viewport, there is no mechanism to prevent the effect from closing it again on the next trigger. This is a UX bug: users on tablet cannot override auto-collapse without switching to a controlled mode they may not know about.

### SHOULD FIX
- [`frontend/src/components/ui/sidebar.tsx:75-78`] The `useEffect` for tablet auto-collapse fires after the first render. On initial mount at tablet width, the sidebar renders with `defaultOpen` (typically `true`) for one frame, then the effect fires and sets `_open` to `false`, causing a visible flash -- sidebar renders expanded then immediately collapses. The initial state computation in `useState(defaultOpen)` (line 72) should account for the tablet case rather than relying on a post-render effect to correct it.
- [`frontend/src/hooks/use-mobile.ts:205-210`] The `useState` initializer reads `window.innerWidth` directly. In SSR or pre-rendering environments where `window` is undefined (e.g., Next.js server rendering, testing frameworks without a real DOM), this will throw a `ReferenceError`. The old code had the same problem, but since this is a rewrite with a new function signature, it was an opportunity to fix it. A safe default with effect-based reconciliation would be more robust.
- [`frontend/src/hooks/__tests__/use-mobile.test.ts:59-60`] `changeHandlers` is a module-level mutable array shared across all tests, only cleared in `beforeEach`. If vitest runs tests in this file concurrently (which it can do depending on config), handlers from one test leak into another, causing non-deterministic failures. This is fragile test infrastructure.
- [`frontend/src/hooks/__tests__/use-mobile.test.ts:109-113`] The first test in the `useBreakpoint` describe block does NOT call `vi.resetModules()` before the dynamic `import()`, while every subsequent test does. If module caching from a prior test or suite contaminates the import, this test may pass or fail non-deterministically. The setup is inconsistent across tests in the same suite.

### NITPICK
- [`frontend/src/hooks/use-mobile.ts:247-249`] `useIsMobile` now delegates to `useBreakpoint()`, which sets up two `matchMedia` listeners. Legacy callers of `useIsMobile` that only care about the mobile threshold now pay the cost of an extra media query listener they do not need. Minor performance regression for existing consumers.
- [`frontend/src/hooks/__tests__/use-mobile.test.ts:140-146`] The boundary test asserts that exactly 1024px is `"desktop"`, which is correct per the `< TABLET_BREAKPOINT` logic. However, the constant is named `TABLET_BREAKPOINT = 1024`, which a reader might interpret as "1024 is tablet." A clarifying comment in the source noting the exclusive upper bound semantics would reduce ambiguity.

jamescrowley321:
## Review: Acceptance Auditor

### PASS
- [AC-1] `useBreakpoint()` returns `'mobile'` below 768px, `'tablet'` between 768-1024px, `'desktop'` above 1024px — implemented at `frontend/src/hooks/use-mobile.ts:8-43`, tested at `frontend/src/hooks/__tests__/use-mobile.test.ts:66-118` (6 test cases covering all three breakpoints, boundary values at 768 and 1024, and dynamic updates via matchMedia listeners)
- [AC-2] `useIsMobile()` still works (backwards compatible) — implemented at `frontend/src/hooks/use-mobile.ts:45-48` (delegates to `useBreakpoint()`, returns `true` only for `"mobile"`), tested at `frontend/src/hooks/__tests__/use-mobile.test.ts:121-145` (3 test cases including explicit backwards-compatibility check at tablet width returning `false`)
- [AC-3] Sidebar auto-collapses to icon mode (not drawer) at tablet width (768-1024px) — implemented at `frontend/src/components/ui/sidebar.tsx:68-82`. `useBreakpoint()` is called; `isMobile` is derived as `breakpoint === "mobile"`, `isTablet` as `breakpoint === "tablet"`. A `useEffect` sets `_setOpen(false)` when `isTablet` is true, collapsing the sidebar to icon mode. The Sheet/drawer rendering path (line 190) is gated on `isMobile`, so tablet does NOT trigger the drawer. No dedicated integration test exists for the sidebar auto-collapse effect, but the hook behavior that drives it is fully tested.

### FAIL
(none)

### PARTIAL
(none)

### SCOPE CREEP
(none detected -- all changes are traceable to the DS-2.2 ACs)

### Architecture Violations
(none -- changes follow existing patterns: hook in `hooks/`, test in `__tests__/`, sidebar modification uses existing CVA/cn/data-slot conventions, no emoji, no marketing copy)

### Summary
- Total ACs: 3
- Pass: 3 | Fail: 0 | Partial: 0

